Issue #92, opened 2026-07-24
feat(settings): split regular vs admin settings; add Literature-chat & Interface-preference tabs


tricktreat:
### Problem / motivation
`/settings` mixes personal settings and admin-only management in one page (admins see an extra 'Admin' tab group inline). It should be split so the regular settings page is the same for everyone, and admin management lives behind a dedicated entry.

### Area
frontend (settings, app shell)

### Proposed solution
- **Regular `/settings`** — unified for all users. Tabs: Profile, SSH credentials, My LLM, Usage, MCP access, **Literature chat** (new), **Interface preferences** (new). The literature-chat full-text-index toggle and the interface-preferences section are extracted out of the Profile tab into their own tabs.
- **New `/settings/admin`** page — admin only (non-admins redirected to `/settings`). Tabs: LLM admin, Daily papers, Users, Codes, Feedback, Usage overview.
- **Top-right gear icon** in the app shell topbar, shown only to admins, opens `/settings/admin`.

### Alternatives considered
Keeping the inline admin tab group — rejected; the ask is a clean split with a separate admin entry point.

PR #96, opened 2026-07-24
feat(settings): split admin settings out of user settings

tricktreat:
## What

The settings page mixed personal and admin tabs in a single row, separated only by a small "Admin" label — they were never really separated. Now:

- **`/settings`** keeps only personal tabs, and gains two of its own: **Literature chat** (the `chat_fulltext_index` opt-in) and **Interface preferences**, both promoted out of the profile page's inline sections.
  Tabs: Profile / Literature chat / Interface preferences / SSH credentials / My LLM / Usage / MCP access.
- **`/admin`** is a new admin-only page: LLM admin / Daily papers / Users / Codes / Feedback / Usage overview. Non-admins get an empty state with a link back to `/settings` rather than a blank page.
- A **Manage gear button** appears in the topbar for admins, **placed before the language toggle**, with an active state on `/admin`.

## Compatibility

Old `/settings?tab=llm|daily|usage|users|codes|feedback` deep links redirect to `/admin?tab=…`, so existing bookmarks keep working. `/settings?tab=mcp` (and the `/mcp-tools` redirect) is unchanged. Also fixed: the experiment modal's "add an SSH credential" button now deep-links to the SSH tab instead of dropping you on Profile.

The admin tab components stay where they are and are simply exported for the new page — they share a lot of internal helpers with the personal tabs, so moving them wholesale would have been a much larger diff.

`tsc --noEmit` clean; `npm run build` passes.

### DIFF
--- a/src/frontend/src/app/AppShell.tsx
+++ b/src/frontend/src/app/AppShell.tsx
@@ -11,7 +11,7 @@ import { topicPath, useProject } from './project';
 import { SearchPalette } from './SearchPalette';
 import { UserMenu } from './UserMenu';
 import { FeedbackWidget } from '../features/feedback/FeedbackWidget';
-import { api, getToken, type GateDecision, type GateRead, type ReviewMessageRead } from '../lib/api';
+import { api, getToken, isAdmin, type GateDecision, type GateRead, type ReviewMessageRead } from '../lib/api';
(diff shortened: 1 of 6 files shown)


tricktreat:
#92 fixed

PR #167, opened 2026-07-25
fix(desktop): load PDFs in the annotating reader, and stream them by range

### LINKED ISSUE #166: Desktop: the annotating PDF reader fails with "file may be corrupted"
## Symptom

In the desktop client, opening a paper in the **annotating reader** shows `PDF 打不开，文件可能损坏。`, while the **standard reader** displays the very same paper fine. The web build is unaffected.

## Cause

Both readers share one blob built from `api.fetchPaperPdf`, but they consume it differently:

- standard reader → `<iframe src=blob:…pdf>` → checked against **frame-src**, which already lists `blob:`
- annotating reader → pdf.js fetches the blob URL → checked against **connect-src**

`buildCsp` in `src/desktop/src/main/protocol.ts` builds `connect-src` from `'self'` plus the configured server origin. In Chromium `'self'` does not cover `blob:` URLs, so the fetch is refused and react-pdf renders its generic error prop — which claims the file is corrupted regardless of the real failure. `img-src`, `media-src`, `frame-src` and `worker-src` all list `blob:`; `connect-src` is the one that was missed.

## Fix

Add `blob:` to `connect-src`, plus a smoke check so it cannot silently regress.

tricktreat:
Closes #166.

Two fixes to the same reader, one on top of the other.

## 1. Desktop: the annotating reader could not load any PDF

It hands pdf.js a `blob:` URL and pdf.js fetches it, so the load is checked against **connect-src** — and `'self'` does not cover `blob:` in Chromium. The fetch was refused and react-pdf fell back to its generic `error` prop, which read "文件可能损坏" for any failure at all. The standard reader was unaffected because `<iframe src=blob:>` is checked against `frame-src`, which already allowed `blob:`.

`buildCsp` now seeds `connect-src` with `blob:`; the smoke suite asserts it next to the existing worker-src / frame-src checks, and `docs/desktop.md` gains a row in "Settings that look removable and are not".

**Verified on a locally packaged macOS arm64 build against the deployed server:** the paper that previously failed now renders.

## 2. Reading a large PDF blocked on a full download

The reader downloaded the whole file into a blob before rendering anything. A 25MB paper over a link that measures ~0.55MB/s showed nothing for ~45 seconds. The backend was never the bottleneck — the same file serves in 0.087s on the server itself.

- the annotating reader now passes the URL plus the auth header to pdf.js, which fetches only the ranges it needs; `disableAutoFetch` stops it from pulling the remainder in the background
- the standard reader is an `<iframe>` and cannot carry an `Authorization` header, so it still needs the blob — that download is now deferred until the user switches to that mode, and shows its own progress text
- PDF existence now reads `paper.pdf_available` instead of being inferred from a speculative full download that 404s
- CORS exposes `Accept-Ranges` / `Content-Range` / `Content-Length`; they are invisible to cross-origin JS by default, and without them pdf.js cannot detect range support and would keep downloading whole files on desktop
- the error copy no longer asserts the file is corrupted — that wording sent this investigation down the wrong path twice

If a server does not support ranges, pdf.js falls back to a full download, so the behaviour degrades to exactly what it was before.

## Checks

- `test_get_pdf_supports_range` (new): asserts 206, the exact byte slice, `Content-Range`, and `Accept-Ranges: bytes`
- `tests/test_paper_reading.py` — 6 passed
- `ruff check` clean on both touched backend files (3 pre-existing findings elsewhere in `tests/` are untouched)
- frontend `tsc --noEmit` + `vite build` pass
- desktop smoke suite passes in full

Range streaming itself is not yet verified end-to-end in a browser — only the endpoint behaviour and the build are.

### DIFF
--- a/docs/desktop.md
+++ b/docs/desktop.md
@@ -81,6 +81,7 @@ result and must not be recomputed remotely.
 |---|---|
 | `plugins: true` in `window.ts` | The manuscript PDF preview is an `<iframe src=blob:…pdf>` rendered by Chromium's built-in pdfium, not pdf.js. Electron disables plugins by default, so dropping this makes the preview blank |
 | `script-src 'wasm-unsafe-eval'` in the CSP | pdf.js ships openjpeg / qcms as WASM |
+| `blob:` in `connect-src` | The annotating reader hands pdf.js a `blob:` URL and pdf.js fetches it; `'self'` does not cover `blob:` in Chromium. Drop it and only that reader breaks — the standard reader is an `<iframe>` and goes through `frame-src` |
 | `style-src 'unsafe-inline'` in the CSP | CodeMirror's style-mod and KaTeX insert rules at runtime; without it the editor and formula rendering break |
 | `role: 'editMenu'` in `menu.ts` | Without it, Cmd+C/V/A/Z stop working in some controls on macOS |
 | Recreating the window after a server change | The preload injection and the CSP response header are both fixed at document load; `reload()` cannot refresh them |
--- a/src/backend/app/main.py
+++ b/src/backend/app/main.py
@@ -66,6 +66,10 @@ def create_app() -> FastAPI:
         allow_credentials=False,
         allow_methods=["*"],
         allow_headers=["*"],
+        # 跨源下这些响应头默认对 JS 不可见。pdf.js 靠 Accept-Ranges / Content-Range /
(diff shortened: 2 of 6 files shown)


tricktreat:
Verified on a locally packaged macOS arm64 build (`electron-builder --mac --arm64`) against the deployed server: the annotating reader now renders the paper that previously failed with "file may be corrupted". The desktop smoke suite passes, including the new connect-src assertion.

PR #217, opened 2026-07-28
fix(chat): keep in-flight answers alive across view switches and carry citations into later turns

### LINKED ISSUE #193: bug: Opening citation in literature dialogue session causes session interruption
### Summary
When a user opens a citation in a literature dialogue session and switches to the literature window, the ongoing session is interrupted and cannot be resumed.

### Area
literature / wiki

### Steps to reproduce
1. Start a literature dialogue session (route: /lab).
2. At the moment the model begins outputting content in the dialogue, open the citation content.
3. Switch to the literature window (e.g., the main literature view).
4. Observe that the previously running session is interrupted and cannot continue.

### Expected vs actual
**Expected:** The session should continue normally after opening the citation and switching windows.
**Actual:** The session is directly interrupted and cannot be resumed.

### Environment
- App version: 0.1.0
- Environment: production
- Viewport: 1706x1264
- User agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36

### LINKED ISSUE #194: bug: Agent fails to recognize previously cited articles in multi-turn conversations
### Summary
In a multi-turn conversation about literature, the agent cannot identify articles cited in previous turns, preventing further discussion about those articles. The user also asks about the citation logic (RAG or keyword search).

### Area
literature / wiki

### Steps to reproduce
1. Start a conversation about literature and ask about the citation logic (e.g., 'Is the citation done via RAG or keyword search?').
2. In a subsequent turn, refer to an article that was cited in a previous turn and ask about its details (e.g., 'What does this article specifically do?').
3. Observe that the agent does not recognize the article and cannot answer the question.

### Expected vs actual
**Expected:** The agent should remember previously cited articles and be able to discuss them in later turns, providing details about the article's content.

**Actual:** The agent fails to recognize the article from the previous turn and cannot answer further questions about what the article does.

### Environment
- App version: 0.1.0
- Environment: production
- Viewport: 1706x1264, User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36
- Route: /lab
- Research direction: OPSD for Deepresearch Agent

tricktreat:
Fixes #193 and #194. Both live in the literature chat, so they ship together.

> Based on #215; GitHub will retarget this to `main` once that merges.

## #193 — opening a citation and switching views kills the session

`ChatSurface` aborted the SSE on unmount:

```tsx
useEffect(() => () => stopRef.current?.(), []);
```

Opening a citation and switching to the literature view *is* an unmount, so the answer was cut off mid-stream. Worse, the last assistant message was left at `done: false`, so coming back showed a **spinner that never stops** — no content, no way to retry.

In-flight streams now live in a module-level registry (`liveChatStreams`), owned by the conversation rather than the component:

- **Unmounting only unsubscribes; the stream keeps running.**
- While nobody is subscribed, the registry **persists progress to localStorage** (throttled to 400ms, and mutually exclusive with the mounted component's `commit` so the two never fight over the same key).
- **Coming back re-subscribes and immediately receives the full snapshot**, then follows the rest of the stream to completion.
- Pressing stop still genuinely aborts the connection and never delivers a partial answer downstream.

This also required splitting the storage layer out of `useChatHistory` into `chatStorage` (no React import) — a stream that outlives the component cannot persist through a hook.

## #194 — later turns cannot identify papers cited earlier

Worse than the report suggests. Context is retrieved and **renumbered `[1]..[n]` on every turn**, while history carried only `role` and `content`:

```js
.map((m) => ({ role: m.role, content: m.content }))
```

So the model neither knows which paper the previous turn's `[1]` was, nor that it differs from *this* turn's `[1]`. It does not just forget — it confidently answers about **the wrong paper**.

- Assistant turns in history now carry `cited_paper_ids`, and the backend appends a "citation key" line to that turn explaining its own numbering. **Titles are read from the database, not trusted from the client.**
- The system prompt states that numbering does not carry across turns, and that an unqualified "this paper" refers to the most recent citation key.
- **Papers cited in the last two turns that this turn's retrieval missed are pinned into context** (restricted to the current scope). This is the part that actually makes follow-ups answerable: "what exactly does this paper do" carries almost no semantic signal, so retrieval will not bring that paper back, and the model would recognise the title while having no text to reason from.

All three entry points — project library, single library, and shelf/personal library — share one implementation.

## Verification

- Two new backend tests, both confirmed to **fail without the fix** (verified by temporarily reverting to the old logic): one pins the citation key on history turns, the other builds a "retrieval only hits the other paper" scenario and asserts the previously cited paper is still pinned into context, ordered after the retrieved results.
- 47 chat-related backend tests pass; `ruff check app` clean.
- #193 is pure frontend interaction and the repo has no frontend test framework. Making `liveChatStreams` / `chatStorage` React-free allowed bundling them with esbuild and running the full scenario under node: mounted receives deltas → after unmount the stream is still alive and content keeps persisting → re-subscribing yields the full text → completion pushes done → an unattended stream still persists its terminal state → stop genuinely aborts. All pass.
- `tsc --noEmit` and `vite build` pass.

Closes #193
Closes #194

### DIFF
--- a/src/backend/app/api/daily.py
+++ b/src/backend/app/api/daily.py
@@ -434,7 +434,7 @@ async def chat_with_daily_pool(
     """
     from app.api.wiki import _chat_stream_response
 
-    history = [(turn.role, turn.content) for turn in data.history[-20:]]  # 最多 10 轮
+    history = library_chat_service.history_from_turns(data.history[-20:])  # 最多 10 轮
     paper_ids = await daily_service.daily_paper_ids(session)
     messages, sources = await library_chat_service.build_scoped_messages(
         session,
--- a/src/backend/app/api/libraries.py
+++ b/src/backend/app/api/libraries.py
@@ -988,7 +988,7 @@ async def chat_with_library(
     library = await _get_visible_library(session, library_id, user)
     user_id = user.id  # 先快照：检索失败路径的 rollback 会使 ORM 对象过期
     project_id = library.project_id
-    history = [(turn.role, turn.content) for turn in data.history[-20:]]  # 最多 10 轮
+    history = library_chat_service.history_from_turns(data.history[-20:])  # 最多 10 轮
     llm = get_llm_router()
     messages, sources = await library_chat_service.build_library_messages_for_library(
         session, library=library, question=data.question, history=history, llm=llm, user_id=user_id
--- a/src/backend/app/api/wiki.py
+++ b/src/backend/app/api/wiki.py
@@ -270,7 +270,7 @@ async def chat_with_library(
     """
     project = await _member_project(session, project_id, user)
     user_id = user.id  # 先快照：检索失败路径的 rollback 会使 ORM 对象过期
-    history = [(turn.role, turn.content) for turn in data.history[-20:]]  # 最多 10 轮
+    history = library_chat_service.history_from_turns(data.history[-20:])  # 最多 10 轮
     llm = get_llm_router()
     messages, sources = await library_chat_service.build_library_messages(
         session, project=project, question=data.question, history=history, llm=llm, user_id=user_id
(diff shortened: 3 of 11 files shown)


tricktreat:
分支已 rebase 到合并后的 main（#215 已 squash 进去），改基后重新开在 #223。本 PR 因基线分支删除被自动关闭，内容与 #223 一致。

Issue #216, opened 2026-07-28
Library sync never runs: daily_wiki_ingest is not registered with the worker


tricktreat:
## Symptom

No library has synced since the switch to event-driven syncing. The daily fetch itself succeeds — `daily_feed_sync` for 2026-07-28 finished `done` and the pool filled up (435 entries for 07-28, 225 for 07-27) — but no scoring, ingest or wiki compilation follows. `voyage_runs` contains no `wiki_ingest` row for today at all, and nothing surfaces in the UI: the fetch reports success, so the failure is invisible.

Enqueueing the job by hand makes it obvious:

```
10:14:58: job 5dce9ce9631d492db7ef730888c53a2e, function 'daily_wiki_ingest' not found
```

## Cause

`daily.sync_libraries`, the last step of the daily fetch, does:

```python
queue = await get_task_queue()
await queue.enqueue("daily_wiki_ingest")
```

But `WorkerSettings.functions` in `src/backend/worker/settings.py` does not list it:

```python
functions = [
    ping_task,
    func(run_voyage, timeout=VOYAGE_JOB_TIMEOUT_SECONDS),
    func(resume_voyage, timeout=VOYAGE_JOB_TIMEOUT_SECONDS),
    match_user_publications,
    index_papers_fulltext_task,
    daily_feed_sync,
]
```

arq only executes jobs whose function is registered, so every enqueue is dropped.

`daily_wiki_ingest` used to reach the worker through `cron_jobs`. When #206 / #212 moved library sync from a fixed schedule to being triggered once the pool is ready, the cron entry was removed and nothing added it to `functions`. The worker banner shows the regression directly:

- 2026-07-25: `Starting worker for 9 functions: … daily_feed_sync, cron:daily_feed_sync, cron:daily_wiki_ingest, cron:daily_publication_match`
- 2026-07-28: `Starting worker for 8 functions: … daily_feed_sync, cron:daily_feed_sync, cron:daily_publication_match`

## Fix

Add `daily_wiki_ingest` to `WorkerSettings.functions`.

Worth considering alongside it: `daily.sync_libraries` reports `{"triggered": True}` and logs "已触发各文献库同步" without checking that the job was accepted, which is why a dropped enqueue looked like success for days. Having the step verify the enqueue — or having the fetch voyage surface the resulting ingest runs — would make the next such breakage visible.

PR #222, opened 2026-07-28
feat(daily,papers): library filter + today's sync status, a "new today" view, and one shared detail layout

### LINKED ISSUE #218: feat: add literature library filtering to Daily Papers
### Problem / motivation
The Daily Papers page currently combines papers from all literature libraries into a single list. Users cannot tell which library each paper belongs to or filter the list by library. As the number of papers grows, this makes the page difficult to navigate and the relevant papers hard to identify.

Context: `/daily` route, production environment, app version `0.1.0`.

### Area
frontend

### Proposed solution
- Display the source literature library for each paper on the Daily Papers page.
- Add a literature library filter that allows users to show papers from one or more selected libraries.
- Keep an option to view papers from all libraries.

### Alternatives considered
No alternatives were provided in the feedback.

tricktreat:
Closes #218.

> Based on #215; GitHub will retarget this to `main` once that merges.

## Which library collected this paper — and let me filter by it (#218)

`GET /daily/papers` takes `library_id`, on both the keyword and semantic paths. It matches only real membership: a paper that is still a candidate, or sitting in the recycle bin, does not count as "this library collected it". The advanced-search panel gets a **Library** dropdown, listing only libraries the requester can see.

The other half of #218 — showing the collecting libraries on a paper — ships in #215 as the badge in the detail pane.

## Today's sync, visible without opening five task pages

A fetch that succeeded but collected nothing looks exactly like a fetch that never triggered a sync at all. That is precisely how #216 stayed hidden.

`GET /lab/daily-ingest` returns the funnel for each visible library's most recent ingest today — scanned → scored → kept/dropped → compiled — read out of that run's step observations. The lab workbench's Daily Papers card renders one row per library: status dot, name, and `scanned → kept`. The full funnel is in the tooltip, and the row links to the run. It polls every 15s while any library is still running and stops once they all finish.

The card's hard-coded "last 7 days kept" now follows the configured retention, same bug class the daily page already had fixed.

## "New today" in the library paper list

A fourth view after All / Compiled / Starred, showing only papers auto-collected from the daily feed today. Both conditions matter — **added today AND from the daily pool** — so papers added by hand or pulled in by citation expansion do not appear. The header row states when the batch arrived.

Backend takes a `daily_only` flag (default `false`, so citations export and the shelf are untouched) alongside the existing `created_from`.

## One layout across all five detail panes

Daily papers, library workbench, read-only library browse, personal library, and related-research shelf now render in the same order:

```
我的标签 → 概念（+ 课题备注·为什么相关）→ 元信息 → 摘要
→ TL;DR → 我的笔记 → 重要图片 → AI 图文介绍
```

Previously concepts sat *after* the metadata block, the abstract fold was hand-written three separate times with different styling, and in two panes the abstract was not collapsible at all. Metadata, abstract and notes now all use `MetaFold`, restyled to match the "My notes" card, all collapsed by default. Missing fields drop the whole block instead of leaving an empty "no abstract yet" box.

### One bug this would have introduced

`MetaFold` keeps its open/closed state internally, and the library workbench and browse panes do **not** remount when you switch papers (the other three are keyed at the call site). Expanding an abstract would have left the next paper's abstract expanded too. Both now pass `key={paperId}` to their folds, replacing the manual `setAbstractOpen(false)` reset those panes used to carry.

## Verification

- Backend suite: **888 passed**. `ruff check app` clean. `tsc --noEmit` and `vite build` pass.
- Six new tests: the library filter includes collected papers and excludes candidates; an unknown library id returns nothing; the ingest funnel reports the right numbers from step observations; a personal library's status is visible to its owner and hidden from a stranger (the earlier version of this test was passing vacuously — a project's implicit library is public, so it proved nothing; it now creates a real personal library); the status list is empty with no runs; and "new today" excludes both manually added papers and papers collected yesterday.

### DIFF
--- a/src/backend/app/api/daily.py
+++ b/src/backend/app/api/daily.py
@@ -86,6 +86,8 @@ async def list_papers(
     category: str | None = Query(default=None, max_length=32),
     author: str | None = Query(default=None, max_length=200),
     affiliation: str | None = Query(default=None, max_length=200),
+    # 只看被某个文献库收录的（每日论文页按库筛选）
+    library_id: uuid.UUID | None = Query(default=None),
     session: AsyncSession = Depends(get_session),
     user: User = Depends(current_active_user),
 ) -> DailyPage:
@@ -113,6 +115,7 @@ async def list_papers(
                 announce=announce,
                 author=author,
                 affiliation=affiliation,
+                library_id=library_id,
             )
             mode_used = "semantic"
             entry_by_paper = {paper.id: entry for entry, paper, _ in rows}
@@ -143,6 +146,7 @@ async def list_papers(
             category=category,
             author=author,
             affiliation=affiliation,
+            library_id=library_id,
         )
         return DailyPage(items=items, total=total, page=page, size=size, mode_used=mode_used)
     ready, ready_total = await daily_service.embedding_coverage(session)
--- a/src/backend/app/api/lab.py
+++ b/src/backend/app/api/lab.py
@@ -13,7 +13,12 @@ from app.api.auth import current_active_user
 from app.core.db import get_session
 from app.models.user import User
 from app.schemas.graph import GraphResponse
-from app.schemas.lab import LabLeaderboardEntry, LabLeaderboardResponse, LabStats
+from app.schemas.lab import (
+    DailyIngestStatusRow,
+    LabLeaderboardEntry,
+    LabLeaderboardResponse,
+    LabStats,
+)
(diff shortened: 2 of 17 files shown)


tricktreat:
分支已 rebase 到合并后的 main（#215 已 squash 进去），改基后重新开在 #224。本 PR 因基线分支删除被自动关闭，内容与 #224 一致。

PR #303, opened 2026-08-06
fix(library): adding a paper shouldn't die because arXiv is throttling us

tricktreat:
Manual add failed with "Internal Server Error". The traceback from the deployment:

```
ArxivRateLimitedError: arXiv unavailable after 4 attempts (HTTPStatusError)
httpx.HTTPStatusError: 429 from arXiv
```

arXiv was rate-limiting us, the exception reached the endpoint unhandled, and the user got five words that say neither what happened nor whether retrying would help.

## Fall back to OpenAlex

The same paper's metadata is in OpenAlex. The resulting record is thinner — abstract often missing, no primary category — but **a partial record beats no record**. The user is holding that arXiv id; making them wait on someone else's rate limiter is passing another service's outage on to them.

## When both sources are down, say so

Upstream throttling is now **503 `UPSTREAM_RATE_LIMITED`** instead of 500, and the dialog explains that arXiv is throttling us, suggests waiting a few minutes, and points at DOI / BibTeX as a way through right now.

"Failed to add" reads as "you typed the id wrong" — it sends the user hunting for a mistake they didn't make.

## Tests

Two: a throttled arXiv resolves through OpenAlex with the id preserved, and both sources failing raises a readable `ParseFailedError` rather than escaping to the endpoint.

The `test_paper_manual_add.py` suite passes (12). The full suite was still running when this was opened — I'll report the number.

### DIFF
--- a/src/backend/app/api/libraries.py
+++ b/src/backend/app/api/libraries.py
@@ -86,6 +86,7 @@ from app.services import papers as papers_service
 from app.services import research_digest as research_digest_service
 from app.services import statement_interview as interview
 from app.services.embedding import embed_query
+from app.services.literature.arxiv import ArxivRateLimitedError
 from app.services.wiki_export import build_obsidian_zip_for_libraries
 
 router = APIRouter(tags=["libraries"])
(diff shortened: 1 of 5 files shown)


tricktreat:
Full suite: **1137 passed**. The single failure (`test_debug_limit_terminates`) pre-exists on `main` and is unrelated — it reproduces with these changes stashed.

PR #359, opened 2026-08-07
feat(lab): count library and paper views, and show a 7-day heat list

tricktreat:
Clicks on a library or a paper are recorded, and the lab workbench gets two seven-day lists: which libraries and which papers people actually opened.

## Three decisions that decide whether the numbers mean anything

**One person's repeated clicks count once an hour.** Without deduplication the chart measures who refreshes most eagerly — switching tabs on a single paper would top it. An hour is the scale of "reading a paper": re-opening from another entry point, refreshing, coming back after a detour are all the same read.

**Deleted objects leave the chart but keep their history.** The event table deliberately has no foreign key to the target: "this was read" stays true after the target is gone. Queries filter by what is currently visible instead.

**Only totals are returned, never per-person rows.** In a lab, "who is reading what" is sensitive; heat needs the count and nothing more. Libraries are additionally filtered by visibility — a library you can't see must not appear, however hot it is.

## Shape

Its own table rather than `activities`: that stream is semantic events, each carrying a human-readable message for the timeline, and high-frequency clicks would flood it.

Recording is best-effort and fire-and-forget from the frontend (202, failures swallowed). A paper page that won't open because analytics broke is far worse than a missing chart. Empty lists render nothing at all — a chart of zeroes carries no information and reads as a broken feature.

## Also here: the status badge was lying

Papers that had already been compiled still showed **"included"** in lists. The badge rendered the membership row's `status`, while "has this been compiled" is a paper-level fact stored in `paper_wikis`. A manually included paper stays `included` forever, compiled or not — so the badge could never say otherwise.

The badge now follows the fact (`has_wiki`) rather than the field, in the library browser, the papers tab, and the reading panel.

## Tests

Six on the counting rules: dedup per person, dedup is per-person (two people each count), the hour boundary, unknown kinds refused, visibility filtering (verified with a non-admin — admins see every library by design, which would have made the assertion vacuous), and a counts-only payload. The migration round-trip is extended for the new table.

Frontend: vitest **48 passed**, `tsc --noEmit` and `vite build` clean. Full backend suite was still running when this was opened — I'll report the number.

### DIFF
--- a/src/backend/alembic/versions/a1c9e73b5d20_view_events.py
+++ b/src/backend/alembic/versions/a1c9e73b5d20_view_events.py
@@ -1,0 +1,43 @@
+"""view events
+
+Revision ID: a1c9e73b5d20
+Revises: 63133f647463
+Create Date: 2026-08-06
+
+文献库与论文的浏览事件。目标 id 不加外键：删掉的对象不该把「它曾被读过」这件事
+一并抹掉，查询时按当前可见对象过滤就够了。
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "a1c9e73b5d20"
+down_revision: str | None = "63133f647463"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "view_events",
+        sa.Column("id", sa.Uuid(), primary_key=True),
+        sa.Column("kind", sa.String(length=16), nullable=False),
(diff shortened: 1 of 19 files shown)


tricktreat:
Full suite: **1206 passed**. The single failure (`test_experiment_env.py::test_prompt_with_context_unit`) pre-exists on `main` — reproduced with these changes stashed.

PR #395, opened 2026-08-12
fix(pipeline): close stage-harness gaps found by two end-to-end runs

### LINKED ISSUE #394: Stage-harness gaps from two end-to-end pipeline runs: template demo leaks into papers, container-blind preflight, premature experiment done, exact-name metric matching, re-gated user extensions, opaque all-unreliable reviews
Findings from two full end-to-end pipeline runs (forge → tournament → experiment → writing → paper review), each observed live in production:

## Writing: template demo content leaks into drafted papers
The draft voyage relies on the client calling `initialize-structure` before `POST /draft`. Skip it (any API-driven flow) and sections fall back into the template's own main file: the demo body — including the template's example citations (`\cite{langley00}`, `\cite{anonymous}`) and example numbers — survives into the compiled paper. Live case: the paper review stage flagged "8 fabricated citations" that were all ICML template samples. Fix: `create_writing_voyage` auto-initializes the structure when the compile main file has no `POLARIS_SECTION` markers.

## Experiment: preflight is container-blind
Resource preflight probes host paths, but container plans declare in-container paths per `plan.container.mounts` (`~/hf → /hf`). `/hf/model/...` is probed on the host, fails, and interrupts the user with a false "resource missing" ask. Fix: reverse-map through the mounts table before probing.

## Experiment: report step still wrote a premature terminal status
#367 removed the premature `failed`; the `done` side had the same structure — report wrote `done` on a successful last run, then done-criteria failed, and the `waiting_user` mirror was blocked by the terminal guard (experiment displayed done while the voyage sat asking). Report now writes no terminal status at all; `done` comes solely from the voyage-done linkage, which also emits the `experiment.status` WS event the report used to send.

## Experiment: exact-name primary-metric matching
Generated code logs metrics with condition/split suffixes (`gsm8k_accuracy/baseline/val`); the plan's primary metric says `accuracy`. Exact-name lookup counted successful runs as zero and dead-ended done-criteria. Matching now normalizes: exact → base-name equality → base-name containment (shortest key wins).

## Engine: user-driven extensions re-gated the user
Appended nodes kept LLM-emitted `requires_gate` (including garbage kinds like an action name), so a user who just decided "continue" was asked to approve consecutive gates for their own decision. The append normalization now strips `requires_gate`.

## Review: all-unreliable aggregation reads as broken scores
When every reviewer fails the guardrail, the summary shows `rating 0.0` with an empty ratings list — indistinguishable from a scoring bug. The aggregation payload now carries `all_unreliable` so the UI can say "no reliable reviews" instead of zeros.

Also observed, not yet fixed (backlog): live tournament comments can land in a session the debate never consumes (stranded human feedback, observed in both runs); container GPU spec frozen at plan time; analyze-ask retry semantics.

tricktreat:
Closes #394

Six fixes across four pipeline stages, each reproduced live during the two end-to-end runs (details in the issue):

1. **Writing — template demo leak**: `create_writing_voyage` auto-runs structure initialization when the compile main file lacks `POLARIS_SECTION` markers. The live paper's "8 fabricated citations" were all ICML template samples surviving a draft that skipped `initialize-structure`.
2. **Experiment — container-blind preflight**: `_host_path_for` reverse-maps in-container refs (`/hf/...`) through `plan.container.mounts` before host probing.
3. **Experiment — premature `done`**: the report step no longer writes any terminal status (mirror of #367's premature `failed`); the voyage-done linkage finalizes and now emits the `experiment.status` WS event.
4. **Experiment — metric-name matching**: `extract_primary_value` matches exact → base-name equality → containment (`gsm8k_accuracy/baseline/val` counts for primary metric `accuracy`; shortest key preferred).
5. **Engine — extension re-gating**: user-driven appended nodes get `requires_gate` stripped (LLM emitted garbage kinds like `experiment.run` and re-gated the user's own decision, twice).
6. **Review — opaque all-unreliable**: aggregation payload carries `all_unreliable`.

## Tests
- New: draft auto-initialization (demo main → draft.tex skeleton, `langley00` gone); `_host_path_for` mapping edge cases; metric normalization precedence; extension node with garbage `requires_gate` passes ungated.
- Suites: experiments / voyage ask / loop / engine / writing / manuscripts / paper review — 117 passed.

### DIFF
--- a/src/backend/app/agents/voyage/actions_experiment.py
+++ b/src/backend/app/agents/voyage/actions_experiment.py
@@ -1142,13 +1142,37 @@ def merge_metrics(target: dict[str, Any] | None, points: list[dict[str, Any]]) -
     return merged
 
 
+def _metric_base(key: str) -> str:
+    """指标键 → 归一化基名：剥掉 /条件/切片 后缀、统一小写。"""
+    return key.split("/", 1)[0].strip().lower()
+
+
 def extract_primary_value(metrics: dict[str, Any] | None, metric_name: str) -> float | None:
-    """从 run.metrics 取主指标最后一个值（无该指标返回 None）。"""
-    series = (metrics or {}).get(metric_name)
-    if not isinstance(series, list) or not series:
-        return None
-    value = series[-1].get("value") if isinstance(series[-1], dict) else None
-    return float(value) if isinstance(value, int | float) else None
+    """从 run.metrics 取主指标最后一个值（无该指标返回 None）。
+
+    匹配做归一化（#20，线上实测）：生成代码打的指标名带条件/切片后缀
+    （``gsm8k_accuracy/baseline/val``），计划主指标叫 ``accuracy``——精确匹配
+    把成功的轮次全计成 0，完成标准死锁。匹配顺序：精确 → 基名相等 →
+    基名包含主指标名（多键命中时取键名最短的，倾向裸指标/baseline）。"""
+    metrics = metrics or {}
+    want = metric_name.strip().lower()
+
+    def last_value(series: Any) -> float | None:
+        if not isinstance(series, list) or not series:
+            return None
+        value = series[-1].get("value") if isinstance(series[-1], dict) else None
+        return float(value) if isinstance(value, int | float) else None
+
+    if metric_name in metrics:
+        return last_value(metrics[metric_name])
+    exact_base = [k for k in metrics if _metric_base(k) == want]
+    containing = [k for k in metrics if want in _metric_base(k)]
+    for candidates in (exact_base, containing):
+        for key in sorted(candidates, key=len):
+            value = last_value(metrics[key])
+            if value is not None:
+                return value
+    return None
 
 
 def is_improvement(value: float, best: float | None, direction: str) -> bool:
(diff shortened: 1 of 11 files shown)


tricktreat:
Also on this branch (9992b4c): **Idea Forge refinement signal** — analysis in #428.

Measured on a hold-out rediscovery bench built on the production OPD library: clone it with only the 106 papers published before 2026-08-12, hide the five later papers, run the real forge pipeline, and have an LLM judge score every generated idea against each hidden paper's core contribution (0-5, where 3 = same mechanism family and 5 = equivalent contribution). Primary target REOPD (2608.11698), whose contribution is making a global reward-extrapolation coefficient token-wise and reliability-adaptive.

| run | change | REOPD | rediscovery@>=3 | idea shape |
| --- | --- | --- | --- | --- |
| baseline | current harness | 3 | 2/5 | 3 of 8 were measure-only programmes |
| ablation | card context + new generation prompt | 2 | 2/5 | 8/8 mechanisms |
| full | + method_knobs signal + interleaving | 3 | 2/5 | 7/7 mechanisms |

The headline numbers did not move, and I am not claiming they did. What did change is measurable: the best REOPD-matching idea grew from 2 matched design elements to 3 (adaptive teacher weight, budget constraint, no extra verifier — the remaining gap is the extrapolation form itself), the whole score distribution shifted up (`3,1,1,1,0,0,0,0` -> `3,2,2,2,1,1,0`), measure-only "systematically evaluate X" ideas disappeared, and the new signal produced 2 of the 7 ideas.

The ablation row is the informative one: the prompt change alone *lowered* the REOPD score, because ideas spread across other gaps while no refinement opening existed to aim at — good prompts without a signal that points at refinement opportunities are not enough.

Full rediscovery of REOPD is capped by the corpus rather than by the harness: the library does not contain ExOPD, the reward-extrapolation paper REOPD builds on, even though REOPD's own abstract cites it. Without that prior the generator can reach "adaptive teacher trust weighting with a budget" but not "amplify the teacher-reference log-ratio".

Changes: new `method_knobs` signal (method cards + deterministic opening aggregation over a controlled knob-category vocabulary, because free-form knob names do not align — 51 cards gave 152 distinct names with one repeat), method-card index in the generation context (covers 60 papers instead of the 20 that fit as wiki excerpts), `builds_on`/`move`/`delta` required per idea with measure-only selections banned, and signal-interleaved gap ordering. Three new tests; forge suite 8 passed.

PR #399, opened 2026-08-12
feat(papers): support replaceable pdf sources

### LINKED ISSUE #398: feat: allow paper PDFs to be uploaded or replaced from a URL
### Problem / motivation

Papers imported from sources other than arXiv may not have a locally available PDF. This prevents the original-paper reader from working and reduces the quality of downstream compilation because full text and figures cannot be extracted. There is currently no way to supply or replace the missing PDF after the paper record has been created.

### Area

literature / wiki

### Proposed solution

Add controls to upload a local PDF or provide a public, directly downloadable PDF URL for an existing paper. Validate the payload, protect URL fetching from private-network access and oversized downloads, replace the stored PDF, invalidate stale extracted text, figures and chunks, and start background enrichment for the replacement.

### Alternatives considered

_No response_

RichardYann:
## Summary

Allow users to upload a PDF or fetch one from a public URL for an existing paper, then rebuild all PDF-derived artifacts so non-arXiv papers remain readable and compilable.

Closes #398

## Area

- [x] frontend
- [x] backend-api
- [x] literature / wiki

## What changed

- Add authenticated endpoints for multipart PDF upload and public PDF URL fetching.
- Validate the PDF header and enforce a 100 MB size limit.
- Protect URL downloads with scheme, credential, DNS/IP, redirect and private-network checks.
- Replace the stored PDF and invalidate stale full text, figures and paper chunks.
- Launch background enrichment after replacement.
- Add upload and URL replacement controls to the PDF reader.
- Add focused regression tests for PDF validation and unsafe URL rejection.

## Testing

- [x] `tsc --noEmit` / frontend production build passes
- [x] Backend tests pass (`tests/test_literature_pdf_sources.py`: 3 passed)
- [x] Ruff passes on all changed backend files

## Checklist

- [x] Branch is rebased on the latest `origin/main` (not merged from main)
- [x] No migration is required
- [x] No AI attribution / `Co-Authored-By` lines in commits
- [ ] Screenshots attached for UI changes

### DIFF
--- a/src/backend/app/api/papers.py
+++ b/src/backend/app/api/papers.py
@@ -10,7 +10,7 @@ from datetime import datetime
 from pathlib import Path
 from typing import Any
 
-from fastapi import APIRouter, Depends, HTTPException, Query, status
+from fastapi import APIRouter, Depends, File, HTTPException, Query, UploadFile, status
 from fastapi.responses import FileResponse, JSONResponse, StreamingResponse
 from redis.asyncio import Redis
 from sqlalchemy.ext.asyncio import AsyncSession
@@ -41,6 +41,7 @@ from app.schemas.paper import (
     PaperMyMetaUpdate,
     PaperMyTagsRead,
     PaperMyTagsUpdate,
+    PaperPdfUrlCreate,
     PaperRead,
     PaperTagsUpdate,
     PaperUpdate,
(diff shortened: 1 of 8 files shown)


tricktreat:
Thanks for this — the SSRF handling here is careful, and re-validating on **every redirect hop** rather than only the initial URL is the part most implementations get wrong.

Two decisions on our side, and one bug, so this needs a re-scope rather than a straight merge.

## Replacement is out

We have decided not to allow replacing a PDF that already exists. #401 landed first and covers the uncontested half — attaching a PDF to a paper that has none — so `POST /papers/{paper_id}/pdf` now exists and returns `409 PDF_ALREADY_EXISTS` when a PDF is present.

This PR and #401 collided head-on: same file, same path, same function name `upload_paper_pdf`, and opposite policy on overwriting. That is our fault, not yours — #398 and #400 were duplicate issues with contradictory scopes, and you two each implemented one faithfully.

What is still wanted from this PR is **fetching a PDF from a URL for a paper that does not have one**. Rebasing onto `main` and dropping the replace path should leave a much smaller diff, and the invalidation logic for full text, figures and chunks is no longer needed since there is nothing stale to invalidate.

## DNS rebinding

`_validate_public_url` resolves the hostname and checks `is_global`, then httpx resolves it **again** when connecting. An attacker controlling the DNS for their own domain returns a public address for the check and `127.0.0.1` for the connection. Since the URL is user-supplied, that is the expected attack rather than an exotic one.

The fix is to pin the address that passed validation and connect to it directly, sending the original `Host` header — or to validate inside a custom transport, where the resolved peer is the address actually used.

## Proxy interaction

The client is built with `proxy=settings.outbound_proxy`. Where a proxy is configured — and it is in our deployment — the proxy performs the DNS resolution and the connection, so the locally computed address does not describe where the request lands. Worth deciding explicitly whether URL fetching should bypass the proxy, so that the validation means what it says.

Happy to look again once it is rebased.

tricktreat:
Update: #398 is now closed by #406.

This PR could no longer be merged. GitHub reports `mergeable=false`, and a trial merge against `main` collides in `api/papers.py`, `services/papers.py`, `PdfReader.tsx` and `lib/api.ts` — the four files #401 changed after this branch was cut. On top of that, replacement is out (that decision predates this), and replacement was the spine of the design here, so what remained was not a small rebase.

Your approach is what shipped, though. #406 reuses it, including the part most implementations get wrong: walking the redirect chain manually and re-validating **every hop** rather than only the initial URL.

Two things were changed:

1. **DNS rebinding.** The address is now re-checked after connecting, before the body is read. The pre-check alone resolves once and httpx resolves again, and a hostile DNS answers differently the second time.
2. **The proxy path.** When `POLARIS_OUTBOUND_PROXY` is set the peer is the proxy, so the peer check is skipped and that is stated in the code, rather than leaving validation that silently means nothing.

Closing this one is up to you — nothing here is wasted, it went in through a different door. Thanks for the SSRF work; the redirect handling was the right instinct.

tricktreat:
Closing this in favour of #406, which shipped the feature and closed #398.

To be clear about why, since the code here was not the problem: this branch was cut before #401 landed, and the two touched the same four files, so it went un-mergeable through no fault of its own. Replacement being ruled out took the rest of it.

The SSRF handling is what carried over — walking redirects manually and re-validating every hop is the part that usually gets skipped, and #406 kept it, adding a post-connect address check and an explicit note about the proxy path.

Thanks for the work here.

PR #413, opened 2026-08-12
feat(papers): add academic provider foundation

atomize-lab:
## Summary

Introduces a provider-neutral foundation for academic discovery and metadata import, building on Polaris's existing arXiv, Semantic Scholar, OpenAlex, import, and resumable-task mechanisms. It also adds Crossref DOI fallback and normalized global paper identifiers so later search-source PRs can share one deduplication and ingestion path.

## Area

- [x] backend-api
- [x] voyage (task loop)
- [x] literature / wiki
- [x] docs

## What changed

- Added typed provider contracts, capability declarations, normalized records, and adapters for arXiv, Semantic Scholar, OpenAlex, and Crossref.
- Added Crossref DOI metadata/relation lookup and fallback when OpenAlex cannot resolve a DOI.
- Preserved the public arXiv ID / DOI / BibTeX import contract while routing it through a typed internal import input.
- Added `paper_identifiers` with normalized globally unique DOI, arXiv, OpenAlex, S2, PMID, and PMCID identities, including legacy backfill and merge handling.
- Migrated arXiv voyage search onto the provider adapter while preserving legacy checkpoints and adding provider-scoped resumable state.
- Reports unsupported provider filters explicitly instead of silently implying they were applied.

## Testing

- [x] Backend tests pass (`55 passed` before rebase; `66 passed` post-rebase including current daily-feed tests)
- [x] Alembic `upgrade head` + downgrade roundtrip passes
- [x] `alembic heads` shows a single head (`8c4a6f1d2e90`)
- [x] `ruff`, `compileall`, and `git diff --check` pass

## Checklist

- [x] Branch is rebased on the latest `upstream/main` (`64956010`)
- [x] New migration uses a random revision id and chains onto the current head
- [x] No AI attribution / `Co-Authored-By` lines in commits
- [x] Screenshots not applicable (no UI changes)


tricktreat:
Reviewed and tested. Holding this one for now — the reason is sequencing, not code quality.

## What was tested

- Migration chains onto the then-current head, single head after merge, upgrade/downgrade round-trip passes
- The backfill was checked against production data: 7721 papers, **zero** duplicate DOIs, zero duplicate arXiv ids, and in `external_ids` arxiv 7613 / s2 602 / doi 561 each with an identical distinct count. So `uq_paper_identifier_value` would not have dropped anything on that deployment
- 170 passed on the branch; 167 passed with `main` merged in, ruff clean across `app/`

The Crossref DOI fallback and reporting unsupported filters explicitly rather than implying they were applied are both worth having.

## Why it is on hold

**The first real consumer arrived and did not use it.** #422 adds Semantic Scholar Corpus ID import — exactly the "later search-source PR" this foundation is for — and it goes through the existing `paper_import.py` path, storing the Corpus ID in the `external_ids` JSON. Moving identifiers out of that JSON into queryable, globally unique rows is this PR's whole purpose.

The two do not conflict textually, which is worse than if they did: both could merge, and the same S2 id would then live in two places with nothing saying which is authoritative. #422 has been merged, so the existing path is where identifiers live today.

## Also worth knowing

Your migration and #427's (`8ff89f7fcdeb`) both set `down_revision = "7b3e91c4a2d8"`. Each is correct alone — both PRs accurately claimed a single head — but merging both produces two alembic heads and `upgrade head` fails, with no git conflict to warn anyone. #427 has since merged, so this branch needs `down_revision` repointed at `8ff89f7fcdeb` whenever it goes forward.

## What would unblock it

A concrete consumer that runs through the abstraction, so the shape gets validated by use rather than by design. Porting #422's Corpus ID path onto `ProviderRecord` + `paper_identifiers` would do it, and would settle where S2 ids live.

One small thing regardless: the backfill skips on identifier collision with no count or log. It changes nothing on this deployment, but a silent drop is the failure mode that has cost us the most time recently — a counter and one warning line would be enough.

Not closing this. The work is sound and the direction is right; it is waiting on a user.

atomize-lab:
Addressed the sequencing feedback and rebased the PR onto the current `main`.

Changes:
- Ported Semantic Scholar Corpus ID import onto `ProviderRecord`.
- Persisted Corpus IDs through `paper_identifiers` using the `s2_corpus` namespace.
- Repointed migration `8c4a6f1d2e90` to `8ff89f7fcdeb`.
- Added collision counting and a warning for skipped backfill identifiers.

Validation:
- 37 targeted tests passed.
- Ruff passed for all files changed by this PR.
- Alembic reports a single head: `8c4a6f1d2e90`.

Ready for another review.

PR #419, opened 2026-08-13
fix(daily): do not spend the day's run on stragglers from an old batch

tricktreat:
Found while answering why nothing from 8-13 had reached a library. The user's instinct — "是不是时区的问题" — was right about the shape of it.

## The timing

The sync fires at `02:00` **UTC**, which is 10:00 Beijing. arXiv announces around 04:00 UTC, 12:00 Beijing. So the run starts two hours before the papers exist.

The probe judged purely on content, so at 10:00 Beijing it found papers from *yesterday's* batch that we had not collected, called that fresh, and ran. `already_ran_today` then closed the day. When the real batch landed at 12:00, nothing went to fetch it.

That is how 2026-08-13 lost 446 papers with every step reporting success: the 02:00 run collected 96 stragglers, and the day was over.

## Two opposite cases

An old batch has to be told apart two ways round, because the right response is the reverse in each:

- **We already have that batch.** The few unseen ones are stragglers, and a run for them costs the entire day.
- **We have none of it** (we were down overnight). This is the last chance — `/new` only ever shows the most recent announcement, so once today's lands, yesterday's is gone.

My first attempt skipped every old batch, which fixed the reported bug and quietly created the second one: one bad night would have become a permanently lost day. The check now asks the pool whether any entry carries that announcement date — answerable because #416 made `feed_date` record the announcement rather than our clock.

Judging by date was abandoned once, and the test that locked that in is still in the suite: the RSS `pubDate` lagged a day, so the check was always false and the pool froze. The listing page states the date in its own header, verified against reality today, so it can be trusted now. Where no date parses, the content-only judgement still applies rather than stalling for a day.

## Testing

- 58 passed in `test_daily_feed`
- Ruff clean
- The straggler test was run against the ungated probe and fails there
- Worth noting: the three older probe tests failed against my first attempt and pass against this one. They were right and the first attempt was wrong — the content-only rule still holds everywhere except the one case this targets.

## Separately

`daily_feed_sync_time` is `02:00` UTC on production. With this fix the probe simply waits and retries until the batch appears, so it is no longer harmful, but setting it nearer 04:00 UTC would save a couple of hours of polling.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01W19Guz3etiapCERw5XbXnr

### DIFF
--- a/src/backend/app/services/daily_feed.py
+++ b/src/backend/app/services/daily_feed.py
@@ -1582,7 +1582,8 @@ async def todays_batch_available(session: AsyncSession) -> tuple[bool, str | Non
     if not categories:
         return True, None
 
-    declared: str | None = None
+    today = _today_utc()
+    latest: dt.date | None = None
     for category in categories:
         try:
             entries, batch_at = await get_arxiv_client().fetch_new(category)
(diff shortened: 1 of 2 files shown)


tricktreat:
Correction to this PR's description — I had the publication timing wrong, and built the explanation on it.

Measured directly:

```
listing page   Showing new listings for Thursday, 13 August 2026
RSS pubDate    Thu, 13 Aug 2026 00:00:00 -0400   = 04:00 UTC
RSS lastBuild  Thu, 13 Aug 2026 04:00:00 +0000
```

I fetched the listing at 02:55 UTC today and it already carried the 13 August batch. **04:00 UTC is when the RSS feed rebuilds, not when arXiv announces.** The listing is out well before our 02:00 UTC sync — the announcement is not late, the feed we were reading was.

So "the sync fires two hours before the papers exist" is wrong. What is true:

- the listing has the day's batch by the time we run
- **RSS only rebuilds at 04:00 UTC**, so a 02:00 UTC run reading RSS always saw yesterday's batch

That also explains why this mostly worked before: the probe would call it not-fresh, keep polling every 15 minutes, and pick it up once RSS caught up at 04:00. The design did cover it. It broke on 08-13 because 08-12 had left 96 uncollected stragglers in that stale batch, so the probe found "papers we have not seen" inside the *old* batch and mistook it for the new one.

The fix in this PR is still right and its mechanism is unchanged — stragglers must not consume the day's single run. Only the account of why the batch was unavailable was wrong.

Worth adding: #415 removes the precondition anyway. Reading the listing page, a 02:00 UTC run gets the current batch immediately and never waits for 04:00. This PR is now defence in depth rather than the primary fix.

The suggestion at the end of the description — moving `daily_feed_sync_time` closer to 04:00 UTC — should be ignored. It was reasoning from the wrong premise; 02:00 UTC is fine, and with the listing source there is nothing to wait for.

PR #441, opened 2026-08-19
fix(writer): preserve ICML frontmatter in structured drafts

### LINKED ISSUE #346: AI-draft skeleton for ICML template omits the title block, making draft.tex uncompilable
## Symptom

A `draft.tex` generated by `initialize_structure` from the ICML 2026 template halts every compile with:

```
! Unbalanced write command.
<write> ...\printAffiliationsAndNotice \string {}
l.NN \end{document}
```

and, because the run dies before BibTeX output is used, every `\cite` additionally surfaces as `Citation 'xxx' undefined`. Reproduced locally with TeX Live and the official `icml2026.sty`: any document that loads `icml2026.sty` but never calls `\twocolumn[\icmltitle{...}...]` + `\printAffiliationsAndNotice{}` dies exactly this way — the style's own end-document reminder uses an unbalanced `\write`.

## Cause

`initialize_structure` keeps the template preamble and replaces the body with the POLARIS_SECTION skeleton. For the ICML template the dropped body included the mandatory title block (`\twocolumn[\icmltitle{...} \begin{icmlauthorlist}...]` and `\printAffiliationsAndNotice{}`), so the generated skeleton can never compile.

## Fix directions

- When building the skeleton, carry over the template's title/author block (everything between `\begin{document}` and the first `\section`/`\begin{abstract}` that is structural rather than prose), or synthesize a minimal one from the manuscript title.
- The diagnostics parser could also translate this specific `Unbalanced write command` into a friendly hint ("ICML template needs \icmltitle + \printAffiliationsAndNotice").

yyy-OPS:
## Summary

Preserve the mandatory ICML title, author, affiliation, and notice block when Polaris replaces the template body with a structured draft. Keep the template bibliography style while consistently targeting Polaris references.bib.

Closes #346

## Area

- [x] backend-api
- [x] writer

## What changed

- detect ICML templates through their package declaration;
- preserve the official frontmatter through printAffiliationsAndNotice;
- replace sample body content with Polaris section markers;
- retain the template bibliography style and use references.bib;
- add a regression test covering the ICML 2026 structure.

## Testing

- [x] pytest tests/test_manuscript_compile_settings.py -q (8 passed)
- [x] ruff check app/services/manuscripts.py tests/test_manuscript_compile_settings.py
- [x] Generated a structured draft from the official ICML 2026 template and compiled it with TeX Live 2025, latexmk, and BibTeX (draft.pdf and draft.bbl produced; no undefined citations)

## Checklist

- [x] No migration

### DIFF
--- a/src/backend/app/services/manuscripts.py
+++ b/src/backend/app/services/manuscripts.py
@@ -430,11 +430,26 @@ _STRUCTURE_SECTIONS: tuple[tuple[str, str], ...] = (
     ("conclusion", "Conclusion"),
 )
 _STRUCTURE_PLACEHOLDER = "To be drafted."
-_BIB_LINE_RE = re.compile(r"^[ \t]*\\bibliography(?:style)?\{[^{}]*\}[ \t]*$", re.MULTILINE)
+_BIB_STYLE_LINE_RE = re.compile(r"^[ \t]*\\bibliographystyle\{[^{}]*\}[ \t]*$", re.MULTILINE)
+_ICML_STYLE_RE = re.compile(r"\\usepackage(?:\[[^\]]*\])?\{icml\d{4}\}")
+_ICML_NOTICE_RE = re.compile(r"\\printAffiliationsAndNotice(?:\[[^\]]*\])?\{[^{}]*\}")
 
 
 def _section_marker_block(key: str) -> str:
     return f"% POLARIS_SECTION: {key}\n{_STRUCTURE_PLACEHOLDER}\n% POLARIS_SECTION_END: {key}\n"
+
+
+def _icml_frontmatter(preamble: str, body: str) -> str | None:
+    """Return the official ICML title block without retaining sample paper content."""
+    if _ICML_STYLE_RE.search(preamble) is None:
+        return None
+    notice = _ICML_NOTICE_RE.search(body)
+    if notice is None:
+        return None
+    prefix = body[: notice.end()].strip()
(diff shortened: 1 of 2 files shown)


tricktreat:
Merged. Reviewed and tested this against the edge cases before merging; the core fix is right, and forcing `\bibliography{references}` is right too — `initialize_structure` guarantees that file exists, so the template's own `\bibliography{example_paper}` points at a file that never will. I verified all three bundled templates (neurips2026 / iclr2026 / acl) produce byte-identical output either way.

Four things surfaced in testing, fixed in #446:

1. `\usepackage{icml2026,times}` was not recognised — `_ICML_STYLE_RE` required the package to sit alone in the braces.
2. `\printAffiliationsAndNotice{\icmlEqualContribution\footnote{...}}` was not matched — `\{[^{}]*\}` cannot span nested braces. Replaced with a balanced-brace scan.
3. In all the miss cases, including a template genuinely lacking the title block, the function returned `None` and the output silently reverted to the uncompilable skeleton — #346 recurring with nothing logged, so the user would only see "compile failed" and start the investigation over. Now warns.
4. The fallback `\bibliographystyle{plainnat}` was appended even when the preamble already declared a style, and BibTeX honours the last declaration — so a template specifying `acm` in the preamble silently compiled as `plainnat`. Now emitted only when neither body nor preamble declares one.

Worth noting for future PRs: (1) and (2) are the kind of thing worth probing directly, since the repo has no ICML template bundled — the regexes were only ever exercised against the one spelling in the test. Thanks for the fix.

PR #443, opened 2026-08-19
fix(literature): normalize unsupported PDF image colorspaces

### LINKED ISSUE #442: PDF figure extraction fails on unsupported image colorspaces
## Symptom

Figure extraction can abort when a PDF contains an embedded image whose PyMuPDF pixmap uses a colorspace that cannot be encoded directly as PNG, such as DeviceCMYK.

## Reproduction

```python
import pymupdf

pix = pymupdf.Pixmap(pymupdf.csCMYK, pymupdf.IRect(0, 0, 10, 10), 0)
pix.tobytes("png")
```

PyMuPDF raises:

```text
ValueError: unsupported colorspace for 'png'
```

The same direct encoding is used by the figure blank-detection and candidate-extraction paths, so one unsupported image can interrupt extraction for the paper.

## Expected behavior

Normalize an otherwise valid pixmap to RGB before PNG encoding when PyMuPDF reports an unsupported colorspace. If an individual candidate still cannot be normalized, skip that candidate with a warning rather than aborting all figure extraction. Other errors should not be silently swallowed.

yyy-OPS:
## Summary

Normalize valid PyMuPDF pixmaps to RGB when their original colorspace cannot be encoded directly as PNG. If an individual figure still cannot be normalized, skip that candidate instead of aborting figure extraction for the paper.

Closes #442

## Area

- [x] backend-api
- [x] literature / wiki

## What changed

- add a focused pixmap-to-PNG normalization helper;
- convert unsupported colorspaces such as DeviceCMYK to RGB;
- keep unrelated encoding errors visible to direct callers;
- treat an unencodable figure candidate as an invalid candidate rather than a paper-level failure;
- add regression coverage for CMYK, RGB, unrelated errors, and candidate failure isolation.

## Testing

- [x] pytest tests/test_pdf_extract_colorspace.py tests/test_paper_figures.py tests/test_figure_crop_expand.py -q (20 passed)
- [x] ruff check app/services/literature/pdf_extract.py tests/test_pdf_extract_colorspace.py

## Checklist

- [x] No migration

### DIFF
--- a/src/backend/app/services/literature/pdf_extract.py
+++ b/src/backend/app/services/literature/pdf_extract.py
@@ -133,9 +133,25 @@ def _white_fraction(img: Any) -> float:
     return sum(1 for v in pixels if v >= 245) / len(pixels)
 
 
+def _pix_png_bytes(pix: Any) -> bytes:
+    """Encode a pixmap as PNG, normalizing colorspaces unsupported by PyMuPDF."""
+    try:
+        return pix.tobytes("png")
+    except ValueError as exc:
+        if "unsupported colorspace" not in str(exc).lower() or pix.colorspace is None:
+            raise
+        import pymupdf
+
+        return pymupdf.Pixmap(pymupdf.csRGB, pix).tobytes("png")
+
+
 def _pix_white_fraction(pix: Any) -> float:
     """Pixmap 铺白后的近白占比（判断 SMask 合并是否把图压成空白 / 候选是否空白）。"""
-    return _white_fraction(_flatten_to_white_rgb(pix.tobytes("png")))
+    try:
+        return _white_fraction(_flatten_to_white_rgb(_pix_png_bytes(pix)))
+    except (RuntimeError, ValueError):
+        logger.warning("cannot normalize PDF pixmap for blank detection", exc_info=True)
+        return 1.0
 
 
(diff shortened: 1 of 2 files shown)


tricktreat:
Merged — note I flipped it out of draft, since the checklist and testing section read as complete; say the word if that was premature.

Verified end to end: a non-blank DeviceCMYK pixmap now yields a valid candidate where it previously raised. Two things I checked that the PR gets right and that are easy to get wrong:

- the `1.0` sentinel from `_pix_white_fraction` is safe at its only call site — unencodable merged pixmap keeps the raw image, unencodable raw keeps the merged one, both unencodable falls through to `_candidate_from_pix` and is dropped with a warning rather than crashing;
- the scoping. The other two `tobytes("png")` sites in the codebase (`actions_review.py`, `presentation.py`) render whole pages via `page.get_pixmap()`, which always yields RGB, so they genuinely do not need this.

One test gap closed in #446: `test_pix_png_bytes_normalizes_cmyk` fills the pixmap with `0`, which in CMYK is no ink — white. A candidate that white is discarded by the blank check before the fix could matter, so the test only ever exercised `_pix_png_bytes`, not the path the issue is about. Added a non-blank CMYK pixmap that runs the whole of `_candidate_from_pix`. Thanks for the fix.

PR #445, opened 2026-08-19
fix(llm): keep provider settings accessible after key rotation

### LINKED ISSUE #444: LLM settings become unavailable after encryption-key rotation
## Symptom

If the Fernet encryption key is rotated while an existing LLM provider still contains an API key encrypted with the previous key, the administrator provider-list endpoint attempts to decrypt the stale value to produce its masked representation. The settings page then fails instead of allowing the administrator to replace the key.

## Expected behavior

The provider list should remain available and expose an explicit masked state such as `*** (needs reconfiguration)` for a value that cannot be decrypted with the current key. Runtime provider construction and model calls must continue to fail closed until the key is reconfigured.

yyy-OPS:
## Summary

Keep the administrator LLM provider settings page usable after a Fernet encryption-key rotation. A stale encrypted API key is shown as needing reconfiguration, while runtime provider construction remains strict and fails closed.

Closes #444

## Area

- [x] backend-api
- [x] llm routing / usage

## What changed

- handle InvalidToken and malformed-token ValueError while producing the masked key;
- return an explicit reconfiguration marker instead of failing the provider list endpoint;
- add regression coverage for stale and malformed encrypted values.

## Testing

- [x] pytest tests/test_admin_llm.py -q (18 passed)
- [x] ruff check app/services/llm_admin.py tests/test_admin_llm.py

## Security note

This change affects only the administrative masked representation. Runtime decryption and model calls still fail closed until the provider key is replaced.

## Checklist

- [x] No migration

### DIFF
--- a/src/backend/app/services/llm_admin.py
+++ b/src/backend/app/services/llm_admin.py
@@ -11,6 +11,7 @@ from collections.abc import Sequence
 from datetime import timedelta
 from typing import Any
 
+from cryptography.fernet import InvalidToken
 from sqlalchemy import delete, func, select
 from sqlalchemy.ext.asyncio import AsyncSession
 
@@ -42,7 +43,12 @@ def mask_api_key(key: str | None) -> str:
 def masked_key_of(provider: LLMProviderConfig) -> str:
     if not provider.api_key_encrypted:
         return ""
-    return mask_api_key(decrypt_secret(provider.api_key_encrypted))
+    try:
+        return mask_api_key(decrypt_secret(provider.api_key_encrypted))
+    except (InvalidToken, ValueError):
+        # Keep the settings page recoverable after an encryption-key rotation.
+        # Runtime provider construction still decrypts strictly and fails closed.
+        return "*** (needs reconfiguration)"
 
 
 def _normalize_user_agent(value: str | None) -> str | None:
--- a/src/backend/tests/test_admin_llm.py
+++ b/src/backend/tests/test_admin_llm.py
@@ -5,6 +5,7 @@ import uuid
 
 import httpx
 import respx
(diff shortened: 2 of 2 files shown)


tricktreat:
Merged. The core fix is right and the security claim checks out — I confirmed `create_provider_client` decrypts through its own `decrypt_secret` call and still fails closed, and that the update path (`if data.api_key:`) means the mask string can never be written back as a key.

One narrowing applied in #446: the `except` also caught `ValueError`. Probing the actual failure modes:

```
rotated key   -> InvalidToken
garbage token -> InvalidToken
Fernet("not-a-real-key") -> ValueError: Fernet key must be 32 url-safe base64-encoded bytes
```

`Fernet.decrypt` raises `InvalidToken` for every bad token, so the only realistic source of `ValueError` is `Fernet(key)` itself — a misconfigured `POLARIS_ENCRYPTION_KEY`. Catching it dressed a deployment error as a per-provider data error: every provider reads `*** (needs reconfiguration)`, and an administrator following that instruction hits the identical `ValueError` from `encrypt_secret` as a 500, with the one useful clue already swallowed. I ran that scenario to confirm.

Also note `test_masked_key_of_keeps_short_or_unrelated_decrypt_errors` — the name says the errors are kept, the assertion says they are masked. When a test name and its assertion disagree, the name is usually recording the intent that got lost. Replaced with two tests: one asserting the misconfigured-key case propagates, and one performing a real encrypt-then-rotate round trip. The existing test proved "catching InvalidToken works", not "rotation raises InvalidToken" — narrowing the clause needed that second fact, and mocking `decrypt_secret` to raise cannot supply it. Thanks for the fix.

PR #463, opened 2026-08-26
feat(literature): add discovery workspace API and permissions

yyy-OPS:
## Summary
- add library-scoped discovery run create, list, inspect, cancel, and delete endpoints
- expose source-attempt progress and paginated candidate filtering/search/sorting
- enforce personal-library isolation and public-library read-only access
- keep the provider adapter boundary unchanged for follow-up integrations

## Authorization
- platform administrators, library creators, and curators can create, cancel, and delete runs
- visible users can inspect runs and candidates
- personal libraries remain private
- public libraries are readable by unrelated users but are not writable by them

## Dependency and merge order
This is intentionally a stacked PR on top of #450 because the API consumes the persisted search-run, hit, source-attempt, and normalized-candidate contracts introduced there. The current branch includes those contract commits so the branch is testable in isolation.

Required merge sequence:
1. Merge #450 first.
2. Rebase this branch onto the resulting `main`.
3. Drop the contract files and migration already merged by #450, leaving only the API, permission service, router registration, and API tests in this PR.
4. Merge this PR.

The final #463 diff must not duplicate #450's migration or model changes.

## Validation
- `tests/test_literature_discovery_contracts.py`
- `tests/test_literature_discovery_api.py`
- `tests/test_library_governance.py`
- `tests/test_library_approval.py`
- Combined targeted result: 34 passed
- `tests/test_migrations.py` literature/head regression: passed on the dependent branch
- Ruff: passed
- Python `compileall`: passed

Closes #453.
Depends on #450.

## Rebase audit update

Rebased on current main after #448/#450/#452. Diff is now API, permission service, router registration, and API tests only; contract schema and migration are inherited. Router is registered once. Validation: 15 discovery API/contract/ranking tests and Ruff passed. First backend PR in stack.

PR #464, opened 2026-08-26
feat(literature): add reusable PDF blobs and library asset grants

### LINKED ISSUE #454: feat(literature): add reusable PDF blobs and library asset grants
## Problem
OA downloads, browser-extension uploads, and user uploads need one content-addressed PDF asset without copying a private paper into another library.

## Scope
- content-addressed PDF blob metadata and integrity verification
- paper asset records with source, checksum, size, and lifecycle state
- library-scoped asset grants with owner/public-library permission rules
- idempotent DOI/identifier reuse across eligible libraries while preventing private-library leakage
- migration, authorization, and persistence tests

## Non-goals
- no MinerU/PyMuPDF parsing or vector generation
- no browser extension changes

yyy-OPS:
Closes #454.

## Scope

Adds content-addressed PDF blobs, paper assets, and library-scoped grants for OA downloads, user uploads, and browser extension archives. The final tree also keeps the discovery API from #463 registered exactly once so the asset endpoints remain observable when this stacked PR is reviewed.

## Dependencies and merge order

This PR is based on #463. Merge #463 first, then rebase this branch and merge #464. The asset migration `c8d9e0f1a2b3` is deliberately linearized after the discovery contracts revision `a7c8d9e0f1b2` already on the current `main`; do not carry a second copy of that migration from an older branch. The branch used by this PR is `yyy-OPS:feat/literature-pdf-assets-v2`.

The earlier revert/reapply commits were branch hygiene while the stack was being assembled. The v2 branch contains the resolved final tree: discovery API files are present through #463, and `app/api/router.py` registers `literature_discovery` and `paper_assets` once each.

## Validation

- `pytest tests/test_literature_discovery_api.py tests/test_paper_assets.py tests/test_migrations.py -q`: 9 passed
- Ruff passed on router, migration, asset API/service/schema/model, and tests
- SQLite migration upgrade/downgrade confirms the single linear head and asset tables

## Rebase audit update

Rebased on cleaned #463. Diff contains reusable PDF blob, PaperAsset, AssetGrant and tests; discovery files are inherited, both routers registered once. Migration c8d9e0f1a2b3 follows a7c8d9e0f1b2. Validation: 9 asset/discovery/migration tests and Ruff passed. Merge #463 first.


tricktreat:
Blocked on a rebase, and there is one thing here I do not want resolved mechanically.

This branch contains a Revert of #463's `feat(literature): add discovery workspace API and permissions`, plus a Reapply and then a Revert of the Reapply. The net effect is that this tree is **missing three files #463 adds**:

- `src/backend/app/api/literature_discovery.py`
- `src/backend/app/services/literature/discovery_runs.py`
- `src/backend/tests/test_literature_discovery_api.py`

and the two branches conflict on `src/backend/app/api/router.py`. Resolving that conflict in favour of this branch leaves the discovery API **in the tree but never registered on the router** — it imports fine, no test fails, and nothing is observable until someone calls the endpoint and gets a 404. That is the kind of failure this repo has been repeatedly bitten by, so it should not be left to whoever resolves the conflict.

Please say in the PR body whether the revert was deliberate. If it was branch hygiene while untangling the stack, rebase onto `main` after #463 lands so the revert never reaches a merge.

Also note `c8d9e0f1a2b3_pdf_asset_grants.py` here declares `down_revision = "a7c8d9e0f1b2"` while #465 / #466 / #467 / #481 ship the same revision id with `down_revision = "8ff89f7fcdeb"`. Same id, two different parents, two different file hashes — details in my comment on #472. Now that #450 is merged, `a7c8d9e0f1b2` is the head, so this branch's version is the one to keep.

yyy-OPS:
Rebase completed and pushed as 51bfcba. This resolves the missing discovery files/router ambiguity: #463 is now an ancestor and both routers are registered once. c8d9e0f1a2b3 now follows a7c8d9e0f1b2. Targeted tests: 9 passed.

tricktreat:
Landed as #490 (merged), rebased onto current main with authorship preserved.

Both hazards I flagged were real and are recorded in #490:

- `router.py` conflicted exactly as predicted — this branch registers `paper_assets` where main registers `literature_discovery`. Resolved as a union; taking either side alone drops the other's routes, and FastAPI does not complain about a router that simply is not registered.
- `schemas/literature_discovery.py` on this branch is a pre-#463 copy missing `SourceAttemptRead`, `SearchHitRead` and `SearchHitPage`. Diffed against main it adds nothing of its own, so it is purely the older state left over by the Revert/Reapply churn. Taking it would have left `api/literature_discovery.py` importing names that no longer exist.

Your `c8d9e0f1a2b3` (chaining from `a7c8d9e0f1b2`) is the version that landed. The variant in #465/#466/#467/#481/#485 chains from `8ff89f7fcdeb` and should be dropped when those are rebased.

PR #465, opened 2026-08-26
feat(literature): add versioned parsing and vector lifecycle

### LINKED ISSUE #455: feat(literature): add content versions parsing and vector lifecycle
## Problem
PDF assets need an observable, retryable MinerU-first parsing lifecycle with a PyMuPDF fallback and persistent full-text vectors.

## Scope
- immutable content versions for MinerU Markdown, images, and PyMuPDF fallback text
- explicit upload, parsing, fallback, ready, and failed states with retry metadata
- page/section/chunk records and document-level plus chunk-level embeddings
- parser provenance and replacement semantics for reparse
- tests for concurrent work, idempotency, fallback, and persistence

## Non-goals
- no reader UI or external model credentials

yyy-OPS:
## Summary

- add immutable paper content versions tied to a specific PDF asset
- persist parser provenance, progress/error state, Markdown/text manifests, pages, and chunks
- provide MinerU-first parser injection with explicit PyMuPDF fallback and retry-safe version replacement
- add version-level and chunk-level embeddings with persisted vector build states
- expose authenticated parse queue, current content-version, and chunk retrieval endpoints
- register the parser task in the ARQ worker

## Scope and lifecycle
A PDF asset remains immutable. Each parse or reparse creates a new `PaperContentVersion`; the previous version is marked non-current instead of being overwritten. This keeps future sentence/paragraph evidence anchors from silently moving when a document is reparsed.

The parser service accepts a MinerU-compatible async adapter. A deployment can wire the existing MinerU Cloud client and key rotation through this interface. When no adapter is configured or MinerU fails, the persisted state records the reason and the explicit fallback path runs through PyMuPDF. If fallback is disabled, the version becomes terminal `failed` with an error code.

Parsed Markdown, plain text, manifest metadata, page/chunk boundaries, and vector states are persisted under the configured data directory. Document and chunk vectors use the active embedding space and are never marked ready until all writes succeed.

## API
- `POST /libraries/{library_id}/papers/{paper_id}/assets/{asset_id}/content-versions`
- `GET /libraries/{library_id}/papers/{paper_id}/content-version`
- `GET /libraries/{library_id}/papers/{paper_id}/content-version/chunks`

## Dependency and merge order
This is a stacked PR based on #464 (`feat/literature-pdf-assets`) and includes #464 in its diff for independent review/testing. It does not depend on #450 or #463 runtime code. Recommended upstream order: #450, clean/rebase #463, #464, then #455. Before merge, rebase onto the actual upstream migration head so Alembic remains single-headed.

## Validation
- content lifecycle, fallback, reparse, and vector persistence tests: 4 passed
- #454 asset API and permission tests: 6 passed
- migration upgrade/rollback: 1 passed
- Ruff and Python compileall: passed
- no credentials, PDFs, ZIPs, caches, or databases included

Closes #455

## Rebase audit update

This branch is stacked on the cleaned #464 tree and contains only versioned parsing/vector lifecycle changes and its migration/tests. The content migration follows c8d9e0f1a2b3; no older duplicate migration definitions are included.

Validation: 11 content/asset/migration tests passed; Ruff passed. Merge #463 -> #464 -> #465.


tricktreat:
Landed as #492 (merged), rebased onto current main with authorship preserved.

Three things had to be resolved, all of them the same underlying cause — this branch predates #489/#490 and carries its own copies:

- its `c8d9e0f1a2b3` chains from `8ff89f7fcdeb` rather than `a7c8d9e0f1b2`, the second of the two contradictory definitions of that revision. Main's version kept.
- `models/__init__.py` and `router.py` drop the `literature_discovery` exports and route registration. Resolved as a union; taking this branch's side would have unregistered the discovery API.
- `tests/test_migrations.py` had lost main's revision constants and assertions, so taking it wholesale produced a red test. Rebuilt from main's version and extended with `PDF_ASSET_REVISION`, the four content tables, and a third downgrade step so the round trip walks the whole chain.

That last one is worth noting for the remaining PRs: `test_migrations.py` is the guard that catches head splits, and each branch shipping its own truncated copy of it is how the two competing `c8d9e0f1a2b3` definitions stayed invisible. When you rebase the rest, please extend main's copy rather than replacing it.

PR #469, opened 2026-08-26
feat(interdisciplinary): add research profiles and linked literature libraries

### LINKED ISSUE #459: feat(interdisciplinary): add research profiles and linked literature libraries
## Problem
Interdisciplinary projects need durable scope, primary/related disciplines, an exclusive cross-disciplinary library, and linked discipline libraries without changing normal projects.

## Scope
- versioned research profile with draft-to-confirmed transition
- project type marker and discipline metadata carried through prompts and workflows
- automatic exclusive cross-disciplinary library plus multiple linked libraries
- discipline-aware query families and evidence provenance
- permission and migration tests for normal and interdisciplinary projects

## Non-goals
- no changes to external source credentials

yyy-OPS:
Closes #459

## Summary
- persist project research mode (`conventional` or `interdisciplinary`)
- add versioned interdisciplinary scope drafts with primary/related domains, core questions, evidence boundary, and validation conditions
- require explicit confirmation before a scope becomes a confirmed research asset
- create exactly one dedicated interdisciplinary literature library on confirmation
- mark the library with `library_kind=interdisciplinary` and durable domain metadata
- reuse existing project membership, library visibility, curator, and paper membership permissions

## API
- `GET /api/projects/{project_id}/interdisciplinary/scope`
- `PUT /api/projects/{project_id}/interdisciplinary/scope`
- `POST /api/projects/{project_id}/interdisciplinary/scope/confirm`

## Scope exclusions
- no LLM provider changes
- no PPT changes
- no author-side changes
- no production deployment or secrets

## Verification
- 2 interdisciplinary API smoke tests passed
- 29 project/library regression tests passed
- Ruff, Python compileall, and `git diff --check` passed
- migration test is blocked on the repository's existing UTF-8 `alembic.ini` under Windows GBK locale; not a migration assertion failure

## Rebase audit update

The branch is rebuilt on current main. The migration parent is now 7c8d9e0f1b2, the dedicated-library claim is enforced by a partial unique index, and created_by uses nullable SET NULL semantics. Confirmation handles a concurrent unique-claim conflict without leaking a 500.

Validation: 3 interdisciplinary/migration tests passed; Ruff and compileall passed. Merge before #470 and #483.


tricktreat:
Merges cleanly and ruff passes, but it cannot be merged as-is: `e9f0a1b2c3d4_interdisciplinary_profiles.py` sets

```python
down_revision: str | None = "fe8a86942dc7"
```

`fe8a86942dc7` (paper pool contract) was not the head when this branched — the head was `8ff89f7fcdeb`. So this forks the migration chain on its own, independently of the literature series: merging it alone gives two heads, and `alembic upgrade head` then fails on a repo that requires exactly one.

Current head on `main` is now `a7c8d9e0f1b2` (#450). Re-point `down_revision` at it, and please add the upgrade/downgrade head assertion that your own merge gates in #472 require — `tests/test_migrations.py` is where the existing ones live, and a single-head assertion there is what would have caught this before review.

Everything else looks reasonable on a read; I will do the real review once the chain is fixed. #470 is clean and small but depends on this, so it waits.

tricktreat:
Re-reviewed. The `down_revision` is unchanged, and there is something in the PR body that needs correcting:

> migration test is blocked on the repository's existing UTF-8 `alembic.ini` under Windows GBK locale; not a migration assertion failure

I ran it in this repository's own Linux test container, on this branch merged into current `main`:

```
tests/test_migrations.py::test_migrations_sqlite_upgrade_head_and_roundtrip FAILED
E  alembic.util.exc.CommandError: Multiple head revisions are present for given
   argument 'head'; please specify a specific target revision...
```

That **is** the migration assertion failing, and it is the only guard in the repository for this defect. It is not a locale problem — there may well also be a GBK/UTF-8 issue reading `alembic.ini` on Windows, but that is a second, separate problem, and it masked this one. When a test that exists specifically to catch X fails on a branch that does X, the failure should be assumed real until reproduced clean somewhere else. Please run the suite in the container (`docker`-based, same as CI) before writing a failure off as environmental.

The fix is still one line — `down_revision = "a7c8d9e0f1b2"`, the current head after #450 merged. Note **#483 already carries the corrected value** for this same revision id, so the two open PRs now define `e9f0a1b2c3d4` with different parents. That is the same shape as the `c8d9e0f1a2b3` split I flagged in #472, so it is a recurring pattern rather than a one-off — worth fixing at the process level, not just per PR.

One substantive point for when the chain is fixed. The body says confirmation creates *exactly one* dedicated library. The handler does guard that with a check-then-insert, but `interdisciplinary_project_id` carries a plain index, not a unique one:

```python
op.create_index("ix_direction_libraries_interdisciplinary_project_id",
                "direction_libraries", ["interdisciplinary_project_id"])
```

So the invariant lives in application code only. Two concurrent confirms — a double-click, or a retried request — both see `None` and both create a library, and the loser is not cleaned up because the FK is `SET NULL`. A partial unique index on `interdisciplinary_project_id WHERE library_kind = 'interdisciplinary'` would make the claim true in the schema rather than by convention.

Minor: `created_by` here uses `ondelete="RESTRICT"`, while comparable columns elsewhere (including `literature_search_runs.created_by` in the just-merged #450) use `SET NULL`. RESTRICT means a user who ever drafted a scope can no longer be deleted. Probably not intended.

yyy-OPS:
Fixed and pushed as bc6fbfe. e9f0a1b2c3d4 now follows a7c8d9e0f1b2; the dedicated library has a partial unique index, created_by is nullable SET NULL, and concurrent confirmation is handled. Migration/API tests: 3 passed; Ruff and compileall passed.

tricktreat:
Landed as #502 (merged) together with #470, rebased onto current main with authorship preserved. Three fixes went in on top:

1. `down_revision` re-pointed from `fe8a86942dc7` to `d1e2f3a4b5c6`. As noted before, the migration test failure was the `Multiple head revisions` assertion, not the GBK locale.

2. **With that fixed, the same test failed again** — `NotImplementedError: No support for ALTER of constraints in SQLite dialect`. The upgrade adds a column carrying a foreign key and the downgrade drops columns; SQLite needs batch mode for both. Converted to `op.batch_alter_table` (matching `d3a7f1c9b2e4` and `f7c2abfe8aeb`) and named the foreign key so batch mode can reflect it. This second defect was invisible while the first existed, because the run died at head resolution before reaching the round trip — which is the practical cost of writing a red test off as environmental.

3. Added the partial unique index on `interdisciplinary_project_id WHERE library_kind = 'interdisciplinary'`, so "exactly one library per project" holds in the schema rather than only in the confirm handler's check-then-insert.

Still open and not changed: `created_by` uses `ondelete="RESTRICT"` where comparable columns use `SET NULL`, so a user who ever drafted a scope can no longer be deleted. It is `nullable=False`, so that is a design call rather than a rebase fix.

PR #468, opened 2026-08-26
feat(extension): bridge Polaris batch archive workflow

yyy-OPS:
Closes #458

## Summary
- add the Polaris browser extension source tree under `integrations/polaris-browser-extension`
- bridge one Polaris push to one local download task containing multiple independent paper items
- connect the extension to the user-scoped Polaris API key and batch archive protocol
- preserve browser-local PDF cache, checksum validation, identity verification, and independent `library_id + paper_id` archive bindings
- keep existing publisher adapters, assisted PDF capture, local download bridge, and Zotero integration available

## Dependency
This PR is stacked on #467 and should be reviewed/merged after #467. The branch contains the #467 history by design; after #467 is merged, GitHub will show only the extension-specific diff.

## Scope exclusions
- no PPT changes
- no author-side changes
- no production deployment or secrets
- no extension UI redesign

## Verification
- `npm run check` (2 passed)
- `node --check` for every extension JavaScript module
- backend batch/archive contract is covered by #467 tests

PR #466, opened 2026-08-26
feat(reading): add versioned evidence anchors

yyy-OPS:
Closes #457

## Summary

Implements Issue #457: a Polaris-owned browser-extension batch archive protocol.

- one push creates one `download_batches` row with independent paper items
- user-scoped API key rotation and revocation; only HMAC hashes are persisted
- claim/lease/heartbeat/cache acknowledgement/retry workflow
- duplicate cached PDFs are skipped without losing each `library_id + paper_id` binding
- direct archive is coordinate-bound, identity-checked, checksum-checked, and idempotent
- successful archive creates a library-scoped `PaperAsset` and a #465 content version, then queues parsing/vectorization
- private/institution-authorized PDFs remain library-scoped and are never widened to public automatically

## Dependency / merge order

This is a stacked PR based on the PDF asset and content lifecycle work:

`#464 -> #465 -> #467`

The branch includes the parent PR files so it can be reviewed and tested independently. Before merging after the parent PRs land, retain only the #467-specific files and migration `e0f1a2b3c4d5` on top of `d9e0f1a2b3c4`.

This PR intentionally does not include browser UI redesign, PPT changes, a separate download-desk service, or author-side baseline changes. The browser extension is only the client of this Polaris API.

## Verification

```text
pytest src/backend/tests/test_download_batch_protocol.py -q
4 passed

pytest src/backend/tests/test_paper_assets.py src/backend/tests/test_paper_content.py src/backend/tests/test_paper_reading.py -q
23 passed

pytest src/backend/tests/test_migrations.py -q
1 passed

ruff check src/backend/app/api/download_client.py src/backend/app/models/download_client.py src/backend/app/schemas/download_client.py src/backend/tests/test_download_batch_protocol.py
All checks passed

python -m compileall -q src/backend/app src/backend/tests
```

## Rebase audit update

This branch is stacked on #465 and contains versioned evidence anchors, reader resolution, and tests. Evidence migration is 5f6a7b8c9d0 with the versioned chunk foreign key; lint-only issues in the original commit were corrected.

Validation: 10 evidence/content/migration tests passed; Ruff passed. Merge #466 after #465.

PR #472, opened 2026-08-26
docs(literature): record implementation status and follow-up PR series

### LINKED ISSUE #471: docs(literature): track implementation status and runtime follow-ups
## Scope
Update the literature evidence RFC after the first implementation series. Record which boundaries are covered by #450, #452, #463-#470 and define the next dependency-ordered runtime and UI PRs.

## Acceptance criteria
- document RFC status and implementation coverage
- distinguish backend contracts from real provider/runtime and frontend work
- define the next Issue/PR sequence for adapters, discovery UI, OA cache promotion, MinerU, reader evidence, and interdisciplinary orchestration
- preserve existing non-goals and migration rules
- no code, credentials, or deployment changes

yyy-OPS:
Closes #471

## Summary
- amend the literature evidence RFC with implementation coverage for #450, #452, and #463-#470
- distinguish backend contracts from provider runtime, main-site frontend, MinerU, reader, OA promotion, and workflow evidence work
- define the dependency-ordered follow-up Issue/PR series and review gates

## Scope
- documentation only
- no code, migrations, deployment changes, credentials, PDFs, ZIPs, or databases

## Validation
- git diff --check passed
- placeholder and sensitive-pattern scan passed

### DIFF
--- a/docs/rfcs/2026-08-26-literature-implementation-status.md
+++ b/docs/rfcs/2026-08-26-literature-implementation-status.md
@@ -35,11 +35,13 @@ reader feature is already production-ready.
 | #468 | Polaris Extension | Browser-side batch bridge, local PDF cache, checksum/identity validation, and batch archive client |
 | #469 | Interdisciplinary profile | Versioned scope draft/confirmation, project research mode, and dedicated cross-disciplinary library |
 | #470 | Interdisciplinary Skill | Builtin skill-market workflow for scope, retrieval, evidence, ideas, experiments, writing, and review |
+| #474 | Discovery runtime | Worker-executed OpenAlex/Semantic Scholar/arXiv adapters, year-window propagation, source isolation, progress snapshots, cross-source deduplication, deterministic ranking, and candidate-only persistence |
 
-These PRs provide backend contracts and one extension client. They do not yet
-provide the Polaris main-site discovery screen, real multi-source runtime
-orchestration, MinerU Cloud production client, or universal evidence injection
-into every AI workflow.
+These PRs provide backend contracts, a worker-executed runtime, and one
+extension client. They do not yet provide the Polaris main-site discovery
+screen, the extended provider/key-health matrix, OA cache promotion, MinerU
+Cloud production client, or universal evidence injection into every AI
+workflow.
 
 ## Required follow-up series
 
@@ -50,24 +52,14 @@ must be rebased and parent files removed from its final diff.
 
 ### Runtime and provider series
 
-1. **Discovery runtime adapters**
-   - Execute persisted query plans through OpenAlex, Semantic Scholar, arXiv,
-     and PubMed.
-   - Preserve requested count separately from internal candidate budget.
-   - Pass start/end years to every capable source.
-   - Record source progress, retries, rate limits, authentication failures, and
-     partial results.
-   - Reuse existing Polaris clients and YFR-derived provider logic through a
-     narrow adapter interface. Do not copy the standalone YFR application.
-
-2. **Extended sources and credential health**
+1. **Extended sources and credential health**
    - Add Crossref, Europe PMC, HAL, CORE, BASE, Unpaywall, and Sciverse
      adapters where credentials and service contracts permit.
    - Support encrypted key pools, rotation, per-source connection tests, and
      source-level capability reporting.
    - Keep missing metadata null; never use `Unknown journal` as a data value.
 
-3. **OA discovery cache and promotion**
+2. **OA discovery cache and promotion**
    - Download every verifiable OA candidate into a temporary, content-addressed
      cache during discovery.
    - Do not parse or vectorize an unpromoted candidate.
@@ -76,14 +68,14 @@ must be rebased and parent files removed from its final diff.
 
 ### Polaris product series
 
-4. **Library discovery workspace**
+3. **Library discovery workspace**
    - Add the discovery entry to the existing library navigation.
    - Show title, abstract, authors, venue, DOI, OA state, source evidence,
      score dimensions, inclusion rationale, progress, history, filters, and
      sorting.
    - Enforce creator/curator write access and public-library read-only access.
 
-5. **Administrator literature settings**
+4. **Administrator literature settings**
    - Add source enablement, query limits, year defaults, scoring rubric,
      provider keys, key-pool editing, and connection-test dialogs.
    - Reuse the existing LLM settings interaction style instead of introducing a
@@ -91,7 +83,7 @@ must be rebased and parent files removed from its final diff.
 
 ### Processing and evidence series
 
-6. **MinerU Cloud production adapter**
+5. **MinerU Cloud production adapter**
    - Implement upload, acceptance, processing, result download, polling,
      multi-key rotation, concurrency limits, retry windows, and persisted
      status transitions.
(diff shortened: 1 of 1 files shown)


tricktreat:
Reviewed today's series in chronological order. **#448, #450 and #452 are merged**; the remaining 15 are blocked on rebases, for a reason worth stating precisely because it is not visible from any individual PR page.

## What is merged

| PR | Verified |
|---|---|
| #448 | RFC, docs only |
| #450 | 13 new tests pass, 30 existing library/migration tests pass, single alembic head, ruff clean. Schema reviewed: FK `ondelete` distinctions are right (run→library CASCADE, `created_by`/`paper_id` SET NULL), `(run_id, dedup_key)` and `(run_id, source)` uniqueness match the dedup contract, indexes match the query shapes, and `downgrade` drops indexes and tables in reverse order |
| #452 | Verified the determinism claim rather than taking it: shuffling the input 5 ways produces identical ordering and identical scores to 9 decimal places, and one DOI seen as `10.1/x` and `https://doi.org/10.1/X` across two sources merges to a single candidate. Requiring `current_year` as a keyword argument instead of calling `date.today()` internally is the right call — this repo's clock is not stable enough to hide that dependency |

## Why the rest cannot be merged in order

Every branch is based on the then-current `origin/main` (`39be7ce`), which follows the rule in this document. But none was rebased after its predecessor, so each one **re-implements its predecessors' commits as fresh commits with different SHAs** rather than building on them. Almost no ancestry relation in your dependency graph actually holds in git — of the pairs I checked, only `#450 ⊂ #474` does.

Measured, not estimated:

- merging all 18 in chronological order: **25 conflicted files across 8 of the 17 merges**, with `c8d9e0f1a2b3_pdf_asset_grants.py` conflicting three separate times
- merging only the non-draft ones: breaks at the **second** merge (#464)
- after #450 landed, #463 still conflicts — though only because it is unrebased. Its copy of `schemas/literature_discovery.py` is a pure +70-line append over the merged version, so the rebase you describe in #463's own body resolves it completely

## The migration chain is the hard blocker

Taking each of the six literature migrations from the earliest PR that introduces it and asking alembic directly:

```
d1e2f3a4b5c6 (head)
e0f1a2b3c4d5 (head)
e5f6a7b8c9d0 (head)
```

**Three heads, two branchpoints.** Worse, the same revision id has two contradictory definitions depending on which branch you read it from:

| revision | in #464 / #476 / #478 / #479 | in #465 / #466 / #467 / #481 |
|---|---|---|
| `c8d9e0f1a2b3` | `down_revision = "a7c8d9e0f1b2"` | `down_revision = "8ff89f7fcdeb"` |
| `d9e0f1a2b3c4` | `down_revision = "d1e2f3a4b5c6"` (#479) | `down_revision = "c8d9e0f1a2b3"` |

No merge resolution can settle that; it needs one linearization chosen deliberately. Now that #450 has landed, the whole chain re-parents onto `a7c8d9e0f1b2`.

Separately, **#469 chains from `fe8a86942dc7`**, which was not the head when it branched — it forks the chain on its own, before any interaction with this series.

## Two things I would not let a mechanical rebase decide

**#464 contains a Revert of #463's "discovery workspace API" commit** (plus a Reapply and a Revert of the Reapply). Its tree is missing three of #463's files: `app/api/literature_discovery.py`, `app/services/literature/discovery_runs.py`, and `tests/test_literature_discovery_api.py`. The two branches also conflict on `app/api/router.py`. A resolution that takes #464's side there leaves the discovery API **present in the tree but never registered on the router** — no import error, no failing test, and nothing observable until someone calls the endpoint. If the revert was intentional, say so in the PR body; if it was branch hygiene, it needs to not reach a conflict resolution.

**#468 is +13122 across 84 files.** Whatever its merge order, that is not reviewable as one unit alongside fifteen siblings.

## Suggested order from here

Rebase onto current `main` one at a time, re-pointing each `down_revision` at the head as it moves, and drop the parent files from each diff as your own gate already requires:

```
#463 -> #464 -> #465 -> #466 -> #467 -> #468
#474 -> #476 -> #477 -> #478 -> #479
#469 (fix down_revision first) -> #470
#481 last (it carries the #466 anchors)
```

Ping me per PR as each is rebased and I will review and merge it. Reviewing them one at a time against a moving main is the only way the migration chain stays single-headed, and it is what your own merge gates in this document ask for.

tricktreat:
Re-reviewed after today's updates. Status has moved backwards, and I want to be precise about why rather than just repeating the last comment.

## Nothing was rebased

All 15 branches are still based on `39be7cebc` — the commit `main` was at *before* #448/#450/#452 merged. The pushes at 16:47 added commits on top; none of them rebased. Two more PRs (#483, #485) were opened on the same stale base.

## The consequence, measured

Last review, several PRs merged cleanly onto `main` individually. Testing each one against **current** `main` now:

- **#469, #470, #472** merge cleanly
- **the other 14 conflict individually**, mostly on `tests/test_migrations.py`, `schemas/literature_discovery.py`, and `models/__init__.py`

This is not a new problem, it is the old one becoming visible: every branch still carries its own copy of the contracts and the `a7c8d9e0f1b2` migration that are now *in* `main`. Each merged PR makes the un-rebased remainder worse, so the cost of deferring the rebases compounds.

## The duplicate-revision pattern is recurring, not a one-off

Last time I flagged `c8d9e0f1a2b3` being defined with two different parents depending on the branch. The same has now happened to a second revision:

| revision | #469 | #483 |
|---|---|---|
| `e9f0a1b2c3d4` | `down_revision = "fe8a86942dc7"` | `down_revision = "a7c8d9e0f1b2"` |

#483 has the correct value. So the corrected knowledge exists in the series but did not propagate back to the PR that introduces the revision. Two open PRs now disagree about the same migration's parent.

Worth fixing at the process level: a revision id should be authored once, in one PR, and every other branch that needs it should get it by rebasing onto the branch that owns it — never by carrying its own copy.

## A failing guard was written off

#469's body says its migration test failure is `alembic.ini` UTF-8 under a Windows GBK locale, "not a migration assertion failure". Run in this repository's Linux container it fails with `Multiple head revisions are present` — precisely the assertion, catching precisely the defect it exists for. Details in my comment on #469.

I raise it here rather than only there because it bears on the merge gates in this document: "migration upgrade/downgrade coverage when schema changes" only protects the repo if a red result blocks the PR. Running the suite in the container before classifying a failure as environmental would be a worthwhile addition to the gate list.

## What I need to merge anything further

Rebase onto current `main`, one PR at a time, re-pointing `down_revision` at the head as it moves and dropping the parent files from each diff — exactly the gate this document already sets out. Ping me per PR and I will review and merge it.

If it would be more useful for me to do the rebases myself, say so and I will — I cannot push to the fork, so I would open the rebased branches in `ZJU-REAL/Polaris` with authorship preserved, and close the fork PRs in favour of those. That is a bigger intervention than I would make unasked while you are still pushing commits, so it is your call.

yyy-OPS:
Stack audit completed. All literature backend branches were rebuilt from current main and force-updated in the fork. Merge order: #463 -> #464 -> #465 -> #466 -> #481 -> #485; #474 -> #476 -> #477 -> #478 -> #479; #467 branches from #465; #469 -> #470; #483 converges #469 + #474. Migration conflicts were resolved by preserving each revision once. #469 uses e9f0a1b2c3d4 after a7c8d9e0f1b2. #479 preserves d9e0f1a2b3c4 and adds unique merge revision fa1b2c3d4e5f for OA/settings plus content heads. Targeted tests and Ruff passed on every rebuilt branch.

PR #481, opened 2026-08-26
feat(ai): expose sentence-level evidence in literature context

yyy-OPS:
Closes #480.

## Summary

- maps retrieved full-text chunks to current sentence anchors
- returns page rectangles and reader URLs in source payloads
- exposes only explicit sentence labels and keeps paper-level fallback
- covers library chat and Wiki/scoped chat through the shared context builder

## Dependencies and merge order

This is a stacked backend PR. Merge the content/evidence chain #464 -> #465 -> #466, then this PR. #479 is a separate MinerU provider lifecycle and is not a prerequisite for the evidence context contract. Rebase onto `main` after predecessors merge and retain only the shared context and focused tests.

## Tests

- `pytest tests/test_ai_evidence_context.py tests/test_library_chat.py tests/test_evidence_anchors.py -q` (17 passed)
- Ruff passed on changed files

## Rebase audit update

This branch is stacked on #466 and contains sentence-level AI evidence context only. It inherits the evidence migration and does not redefine any migration id.

Validation: 7 AI-evidence/reader tests passed; Ruff passed.

PR #485, opened 2026-08-26
feat(ai): query versioned full text with evidence

### LINKED ISSUE #484: feat(ai): use versioned full text in Agent and MCP tools
## Problem

Parsed PDF content and sentence anchors exist in versioned tables, but Agent and MCP tools still read the legacy full-text path and `paper_chunks` index. This leaves MinerU and PyMuPDF content unavailable to downstream research agents. The legacy PostgreSQL keyword query can also apply `DISTINCT` to JSON columns after joining versioned content, which raises `UndefinedFunctionError`.

## Scope

- Generate sentence anchors whenever a content version finishes parsing.
- Read the current authorized content version in `read_fulltext`.
- Search current content chunks and their vectors in `search_chunks`.
- Return anchor IDs, page locators, rectangles, quoted text, and reader URLs.
- Keep the legacy full-text path and `paper_chunks` as compatibility fallbacks.
- Enforce library membership and active asset grants without selecting JSON columns through `DISTINCT`.

## Acceptance criteria

- Reprocessing the same PDF creates anchors for the new current version without changing old anchors.
- Revoking an asset grant removes that versioned content from Agent and MCP search results.
- Semantic search uses `paper_content_chunk_vectors` in the active embedding space and falls back to keyword search when embeddings are unavailable.
- `read_fulltext` paginates the current content version and does not expose an asset outside the caller's library scope.
- Targeted backend tests and Ruff pass.

## Dependencies

This issue builds on #454, #455, #456, and #480.

yyy-OPS:
Closes #484

## Summary

Agent and MCP tools now use the current parsed PDF content instead of stopping at the legacy full-text path or abstract chunk index. Parsing creates sentence anchors as part of the same content lifecycle, and tool results include the stored evidence locator needed by readers and downstream AI output.

## Changes

- Generate chunk, paragraph, and sentence anchors after MinerU or PyMuPDF parsing succeeds.
- Read current versioned content in `read_fulltext`, with pagination and legacy path fallback.
- Search current content and `paper_content_chunk_vectors` in `search_chunks` before consulting the legacy index.
- Return quoted sentences, page ranges, rectangles, anchor IDs, and reader URLs with versioned results.
- Enforce active AssetGrant and library membership checks with correlated `EXISTS` queries.
- Avoid PostgreSQL `DISTINCT` over JSON columns in versioned chunk search.
- Preserve keyword fallback when embedding or either vector index is unavailable.

## Dependency and merge order

This is a stacked backend PR. Merge these PRs first:

1. #464 PDF blob, asset, and grant model
2. #465 parsed content and vector lifecycle
3. #466 evidence anchors and resolver API
4. #481 sentence evidence context for literature chat and Wiki
5. This PR

If an earlier PR is squash-merged, rebase this branch onto the updated `main` before merging it. The functional change introduced by this PR is commit `38d1411`.

## Tests

```text
29 passed, 1 warning
ruff: all checks passed
compileall: passed
git diff --check: passed
```

The warning is the existing SQLite Alembic reflection warning for `direction_libraries.project_id` and does not affect the tested migration.

## Rebase audit update

Stacked on #481. Diff contains versioned full-text Agent/MCP retrieval, AssetGrant filtering, current-version evidence and tests. It inherits content/evidence migrations and provides sentence/chunk/page evidence with paper fallback. Validation: 26 full-text/content/evidence/MCP tests and Ruff passed.


tricktreat:
Landed as #501 (merged), rebased onto current main with authorship preserved.

`services/paper_content.py` needed a hand merge rather than a side: this branch's copy is the pre-MinerU stub that #499 replaced, but it also carries the real change — collecting created chunks and calling `persist_chunk_anchors`. Main's MinerU version kept with that call grafted on, plus an explicit `flush()` so the chunks have ids before the anchors reference them, and `source=parser_name` rather than `version.parser` (that field is assigned a few lines further down, so reading it there would record the previous parser).

`services/evidence.py` showed +324/-11, but all eleven removals are re-wraps — that file was reformatted when #493 landed. Took your version for the genuine +324.

PR #504, opened 2026-08-27
fix(literature): apply administrator settings at runtime

yyy-OPS:
## Summary

Follow-up to #495, #497, and #498.

- apply persisted administrator source, count, year-window, and score defaults when creating discovery runs
- keep provider credentials out of run snapshots and resolve decrypted key pools only inside the worker runtime
- rotate OpenAlex, Semantic Scholar, PubMed, CORE, and Sciverse credentials without rebuilding every request
- fetch PubMed abstracts via EFetch and enforce provider year windows consistently
- make an explicit empty source list disable discovery instead of falling through to query-plan sources

## Merge independence

This PR is based directly on current `main` (`192f019`) and does not depend on another open PR. It contains no migrations.

## Verification

- `27 passed` across administrator settings, discovery API/contracts/ranking, and runtime tests
- Ruff checks passed
- `git diff --check` passed

### DIFF
--- a/src/backend/app/api/admin_settings.py
+++ b/src/backend/app/api/admin_settings.py
@@ -213,7 +213,7 @@ async def set_literature_search_settings(
 ) -> LiteratureSearchSettings:
     try:
         saved = await literature_settings_service.update_settings(
-            session, payload.model_dump(exclude_none=True)
+            session, payload.model_dump(exclude_unset=True)
         )
     except literature_settings_service.InvalidLiteratureSettingError as exc:
         raise HTTPException(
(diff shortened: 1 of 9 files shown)


yyy-OPS:
Fresh verification against origin/main@192f019:

- Ruff: passed for all changed Python files
- Focused tests: 14 passed, 1 warning`n- Diff check: clean

The warning is the existing FastAPI 422 deprecation warning; no runtime-setting test failed.

tricktreat:
Landed as #523 (merged) with authorship preserved, plus one guard test.

This needed no rebase — based on current main, states its merge independence, no migration, clean merge. A marked change from the earlier series; it made review straightforward.

I verified the credential claim behaviourally rather than by reading: posting a run with `provider_keys: {"openalex": ["sk-SECRET-LEAK"]}` and reading the row back. The string appears in neither `source_config`, `query_plan` nor `progress`, and `api/literature_discovery.py` has no `decrypt` call at all.

Added a test for it, because that invariant had none — the existing `provider_keys` coverage is about masking in the admin surface, not the run snapshot. Worth pinning because of where the failure would land: snapshots get read, exported and logged, so a plaintext key in one turns a single decryption point into anywhere-a-run-is-read, and nothing raises. Commenting out the `pop` reds the new test.

PR #505, opened 2026-08-27
fix(interdisciplinary): persist scope suggestions and evidence

yyy-OPS:
## Summary

Follow-up to #502 and #503.

- add a strict JSON tolerant interdisciplinary scope suggestion endpoint with a concrete deterministic fallback
- persist every editable scope as an immutable version, including query channels and evidence balance
- make scope confirmation update the current version and keep dedicated-library creation idempotent under concurrent requests
- preserve top-level retrieval provenance when cross-disciplinary results are reranked

## Merge independence

This PR is based directly on current `main` (`192f019`) and is independent of the runtime settings follow-up. It adds one migration after the current single head (`f0a1b2c3d4e5`) and keeps one head.

## Verification

- `7 passed` across scope parsing/fallback, interdisciplinary APIs/retrieval, and full migration round-trip
- Ruff checks passed
- Alembic reports one head: `f1a2b3c4d5e6`
- `git diff --check` passed


yyy-OPS:
Fresh verification after the import-order follow-up (84738bd):

- Ruff: passed for all changed Python files
- Focused interdisciplinary and migration tests: 7 passed, 1 warning`n- Diff check: clean

The warning is the existing SQLite Alembic foreign-key reflection warning; no migration or scope test failed.

tricktreat:
Landed as #524 (merged) with authorship preserved, plus one fix and two tests.

The `IntegrityError` recovery path added here does not work. `session.rollback()` expires the `project` instance, and the recovery then reads `project.id` twice — a sync attribute access that needs async IO:

```
sqlalchemy.exc.MissingGreenlet: greenlet_spawn has not been called; can't call await_only() here.
```

So the guard written *for* the race is what fails *in* the race: two concurrent confirms give a 500 instead of the intended idempotent 200. Fixed by using the `project_id` path parameter, which is a plain UUID and survives the rollback.

It also could not be tested as written — the unique index condition is identical to the handler's own lookup, so anything that would trip the index gets found by the lookup first and takes the update branch. No externally-driven test can reach that `except`. I extracted the lookup into `_existing_dedicated_library` so a test can stub it to return None once, reproducing the moment both requests believe no library exists. Reverting the fix reds it with `MissingGreenlet`.

Everything else merged as-is: based on current main, migration chains correctly from `f0a1b2c3d4e5`, clean merge, no rebase needed.

PR #507, opened 2026-08-27
fix(literature): preserve MinerU parse assets and versions

### LINKED ISSUE #506: fix(literature): make MinerU parsing and reparsing reliable
## Problem

The current MinerU lifecycle replaces the active content version before a reparse succeeds. A failed reparse can therefore hide usable structured content. The cloud adapter also keeps token rotation and concurrency limits per parser instance, decodes malformed Markdown with replacement characters, and discards figures and tables from result archives.

## Scope

- Share MinerU token rotation and concurrency limits across parser instances in one worker process.
- Retry upload and result download operations separately from the processing poll loop.
- Start the processing deadline after the PDF upload succeeds.
- Persist explicit upload, parsing, polling and result-download states.
- Validate result ZIP paths, file counts, expanded size and UTF-8 Markdown.
- Persist Markdown, figures and tables under the immutable content-version directory.
- Keep the previous current version until the replacement parse succeeds, then switch atomically.
- Preserve PyMuPDF as an explicit fallback after MinerU fails or exhausts retries.

## Non-goals

- Structured-reader frontend changes.
- New content delivery endpoints.
- Deleting historical content versions.
- Changing the embedding model or vector schema.

## Acceptance criteria

- A failed reparse leaves the previous current version readable.
- Successful MinerU results retain Markdown, image and table assets.
- Invalid archives and invalid Markdown encodings fail with stable error codes.
- Configured keys rotate across parse jobs and share the configured concurrency bound.
- Existing PDF asset, evidence anchor, download archive and vector flows continue to pass their focused tests.

yyy-OPS:
## Summary

- share MinerU token rotation and concurrency limits across parser instances
- separate upload, processing poll and result-download status transitions
- validate ZIP paths, expanded size, file count and Markdown encoding
- persist MinerU Markdown, image and table artifacts under the content version
- keep the previous current version until a replacement parse succeeds

## Problem

A reparse currently marks the new version as current before MinerU finishes. If the cloud parse fails, callers can lose access to usable structured content. Result ZIP handling also replaces invalid UTF-8 and keeps only one Markdown file, which hides encoding failures and drops figures and tables.

## Behavior

The parser now reports `mineru_uploading`, `mineru_parsing`, `mineru_polling` and `mineru_downloading_result`. The processing deadline starts after the upload succeeds. Upload and result download calls use bounded retries.

ZIP output accepts Markdown, image and table assets only. It rejects unsafe paths, oversized archives and malformed Markdown. A successful parse stores those artifacts inside the immutable content-version directory.

Creating a reparse attempt no longer retires the active version. The service switches `is_current` only after content, chunks and evidence anchors have been persisted. Failed reparses retain the previous current version. PyMuPDF fallback behavior is unchanged.

## Compatibility

- No database migration.
- No API schema change.
- No frontend change.
- Existing vectorization still starts after parsing through the same worker task.
- The branch starts directly from `origin/main@192f019` and does not depend on #504 or #505.

## Non-goals

- structured reader endpoints or UI
- historical version deletion
- embedding model or vector schema changes

## Verification

```text
python -m ruff check app/services/mineru.py app/services/paper_content.py tests/test_mineru.py tests/test_paper_content.py
All checks passed

python -m pytest -q tests/test_mineru.py tests/test_paper_content.py tests/test_evidence_anchors.py tests/test_download_batch_protocol.py tests/test_paper_assets.py
29 passed, 2 warnings
```

The warnings are the existing FastAPI 422 deprecation and SQLite Alembic foreign-key reflection warnings. A changed-diff sensitive-value scan and `git diff --check` also passed.

Closes #506

### DIFF
--- a/src/backend/app/services/mineru.py
+++ b/src/backend/app/services/mineru.py
@@ -20,6 +20,22 @@ class MineruCloudError(RuntimeError):
     """A MinerU upload, polling, or result conversion failure."""
 
 
+def _safe_request_failure(prefix: str, error: Exception | None) -> MineruCloudError:
+    """Build a stable error without persisting tokens or signed request URLs."""
+
+    if isinstance(error, httpx.HTTPStatusError):
+        detail = f"HTTP_{error.response.status_code}"
+    elif isinstance(error, httpx.TimeoutException):
+        detail = "TIMEOUT"
+    elif isinstance(error, httpx.RequestError):
+        detail = "NETWORK_ERROR"
+    elif error is not None:
+        detail = type(error).__name__.upper()
+    else:
+        detail = "UNKNOWN"
+    return MineruCloudError(f"{prefix}:{detail}")
+
+
 StatusCallback = Callable[[str], Awaitable[None]]
 _SCHEDULERS_LOCK = Lock()
 _MAX_ARCHIVE_FILES = 10_000
@@ -83,12 +99,7 @@ def _result_payload(payload: Any) -> Mapping[str, Any] | None:
 
 def _safe_archive_name(name: str) -> str | None:
     path = PurePosixPath(name.replace("\\", "/"))
-    if (
-        path.is_absolute()
-        or not path.parts
-        or ".." in path.parts
-        or ":" in path.parts[0]
-    ):
+    if path.is_absolute() or not path.parts or ".." in path.parts or ":" in path.parts[0]:
         return None
     return path.as_posix()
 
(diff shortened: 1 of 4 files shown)


yyy-OPS:
Follow-up hardening is now included in `670272b`.

- redacts signed URLs and request details from MinerU failures
- persists Markdown, artifacts, chunks, and anchors as one failure-safe operation
- marks an incomplete replacement version failed, removes its partial directory, and preserves the previous current version

Verification after the update: Ruff passed; MinerU/PDF/assets/anchors focused suite: `31 passed, 2 warnings`.

tricktreat:
Merged as-is — no changes needed, which is worth saying explicitly given the earlier rounds.

I verified the two claims that carry risk rather than taking them on trust:

**The archive guard holds.** `_safe_archive_name` normalises backslashes before parsing, so Windows-style traversal is caught too. Probed it directly:

| input | result |
|---|---|
| `../evil.md` | rejected |
| `a/../../evil.md` | rejected |
| `/etc/evil.md` | rejected |
| `C:\evil.md` | rejected |
| `..\..\evil.md` | rejected |
| `....//evil.md` | `..../evil.md` — correct, that is a directory literally named `....`, not traversal |
| `./ok.md` | `ok.md` |

Plus 512 MiB expanded and 10 000 files as ceilings.

**The is_current invariant is genuinely tested.** `test_failed_reparse_keeps_previous_current_version` and `test_persistence_failure_marks_attempt_failed_and_keeps_previous_current` are not decorative — reverting `is_current=current is None` back to `is_current=True` at creation reds both of them. That is the check I have been asking for on this series, and it is here.

Based on current main, no migration, clean merge. 22 passed across `test_paper_content`, `test_mineru`, `test_paper_assets`, `test_migrations`; ruff clean.

Issue #509, opened 2026-08-27
feat(admin): add literature provider credential pool lifecycle


yyy-OPS:
## Problem

Administrator settings return masked provider keys but update the entire pool at once. A client can accidentally write a mask back as a credential, and individual keys have no stable identity, enable state or health result.

## Scope

- Add stable credential entry IDs and encrypted add, replace, enable, disable and delete operations.
- Add per-entry connection tests and persisted health metadata.
- Keep raw credentials out of GET responses, logs and search-run snapshots.
- Support OpenAlex, Semantic Scholar, PubMed, CORE and Sciverse key pools.
- Store the PubMed contact email separately and mark arXiv as not requiring a key.
- Preserve environment variables as deployment fallbacks.

## Acceptance criteria

Masked values are never accepted as new credentials. Editing one entry cannot replace another entry or the whole pool. Runtime rotation uses enabled entries only, and every provider test returns a stable capability and error contract.

Depends on #504.

yyy-OPS:
Implemented and landed on `main` via #526. The merge commit accidentally referenced unrelated issue #518; this issue is the correct tracker for the literature provider credential-pool lifecycle.

PR #516, opened 2026-08-27
feat(literature): add reproducible multi-query retrieval and model reranking

### LINKED ISSUE #508: feat(literature): add reproducible multi-query retrieval and model reranking
## Problem

The persisted query plan can contain core and coverage queries, but the runtime executes only the first query for each source. It also sends the full candidate budget to every source. The administrator rubric and ranking service use different score keys, and the current relevance fallback is lexical overlap rather than model reranking.

## Scope

- Define one versioned score contract for relevance, evidence quality, impact, novelty and recency.
- Keep OA availability and interdisciplinary coverage outside the scientific quality total.
- Treat `candidate_budget` as the aggregate recall budget and persist per-source, per-query allocations.
- Generate bounded English retrieval queries through the existing LLM router, with deterministic validation and fallback.
- Execute core and coverage queries, normalize source ranks, deduplicate and fuse candidates before reranking.
- Use the configured rerank route for the final relevance pass, with a deterministic fallback when the model is unavailable.
- Persist query, ranking, weight, model and fallback versions in the run snapshot.

## Acceptance criteria

A requested count of 50 produces at most 50 persisted results, while progress reports the separate aggregate candidate budget. Every configured query and year filter is observable. Repeated runs with the same snapshots can reproduce the deterministic stages. Conventional and interdisciplinary libraries share the same candidate DTO.

Depends on #504.

yyy-OPS:
## Summary

- generate and validate English multi-query retrieval plans with bounded retries
- use the same core/coverage query model for regular and interdisciplinary libraries
- separate requested result count from candidate budget and preserve the requested year window
- deduplicate across queries and providers, then fuse candidates with reciprocal-rank fusion
- add model reranking with deterministic fallback and preserve unranked tail candidates
- standardize relevance, evidence quality, impact, novelty, and recency scoring
- report query, retrieval, deduplication, reranking, and final-result progress accurately
- compile arXiv boolean queries without changing the normalized paper contract

## Dependency and merge order

This PR depends on #504. Its branch is based directly on the #504 head commit so that runtime administrator settings are available during retrieval. It is independent of #509 and does not include the credential-pool changes.

Recommended merge order: #504, then this PR. After #504 is merged, the diff should collapse to the retrieval-quality commit only; rebasing can be performed if the maintainer prefers.

## Verification

- focused backend tests: `57 passed`
- Ruff on all changed Python files: passed
- `git diff --check`: passed
- full backend baseline comparison: `1422 passed, 6 failed`; the same six Windows/path/prompt assertion failures reproduce on clean `origin/main` and are unrelated to this change
- sensitive-data scan: no credentials, tokens, environment files, PDFs, archives, databases, or logs included

Closes #508

### DIFF
--- a/src/backend/app/api/admin_settings.py
+++ b/src/backend/app/api/admin_settings.py
@@ -213,7 +213,7 @@ async def set_literature_search_settings(
 ) -> LiteratureSearchSettings:
     try:
         saved = await literature_settings_service.update_settings(
-            session, payload.model_dump(exclude_none=True)
+            session, payload.model_dump(exclude_unset=True)
         )
     except literature_settings_service.InvalidLiteratureSettingError as exc:
         raise HTTPException(
@@ -234,9 +234,11 @@ async def test_literature_provider(
     source = payload.source.strip().lower()
     started = time.perf_counter()
     try:
-        from app.services.literature.runtime import _default_registry
+        from app.services.literature.runtime import build_adapter_registry
 
-        registry = await _default_registry()
+        registry = await build_adapter_registry(
+            await literature_settings_service.get_runtime_settings(session)
+        )
         adapter = registry.get(source)
         if adapter is None:
             raise ValueError(f"unsupported source: {source}")
--- a/src/backend/app/api/literature_discovery.py
+++ b/src/backend/app/api/literature_discovery.py
@@ -34,6 +34,7 @@ from app.schemas.literature_discovery import (
     SourceAttemptRead,
 )
 from app.services import libraries as libraries_service
+from app.services import literature_settings as literature_settings_service
 from app.services.interdisciplinary_retrieval import apply_profile_to_query_plan
 from app.services.literature import discovery_runs, oa_cache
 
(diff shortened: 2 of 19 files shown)


yyy-OPS:
Follow-up hardening is now included in `9826656`.

- snapshots library include/exclude keywords and scoring rubric into each run
- applies the rubric to query generation and model reranking
- compiles provider-specific plain-text queries for OpenAlex, Semantic Scholar, and Crossref
- rejects unusable non-English fallback plans instead of issuing a generic query
- publishes progress as each provider request completes while preserving deterministic fusion order
- prevents duplicate execution of a run already claimed or completed
- attempts OA resolution for DOI-only candidates and redacts provider failures

Verification after the update: Ruff passed; focused retrieval suite: `26 passed`.

tricktreat:
Landed as #525 (merged), rebased onto current main with authorship preserved.

As your body anticipated, #504 landed first (squashed as #523), and that squash is why four files conflicted — the content was fine. Source files resolved in this PR's favour, test files as a union.

One thing I checked rather than assumed: the `provider_keys` stripping added in #523 sits a few lines below one of the conflict hunks, so it was in range of being lost in the resolution. It survives, and I re-ran the sensitivity check afterwards — removing `source_config.pop("provider_keys", None)` still reds `test_provider_keys_never_reach_the_run_snapshot`. A guard is only worth having if it still bites once the next PR merges on top of it.

PR #517, opened 2026-08-27
feat(admin): add literature provider credential pool lifecycle

### LINKED ISSUE #509: feat(admin): add literature provider credential pool lifecycle

yyy-OPS:
## Summary

- give every provider credential a stable UUID and encrypted persisted value
- expose masked credential metadata without returning plaintext secrets
- add administrator-only create, edit, enable/disable, delete, and per-key connection-test operations
- persist per-key health snapshots for administration without leaking credentials into logs or runtime snapshots
- support round-robin selection for multi-key providers, including OpenAlex
- retain compatibility with legacy string-array credential pools
- reject credentials for arXiv because its public API does not require a key
- prevent disabled credential pools from silently falling back to environment credentials

## Dependency and merge order

This PR depends on #504. Its branch is based directly on the #504 head commit because credential selection is consumed by the administrator-settings runtime introduced there. It is independent of #508 and does not include retrieval-quality changes.

Recommended merge order: #504, then #508, then this PR. The #508-before-#509 order minimizes review conflicts in shared runtime/settings files, although the feature commits themselves are independent.

## Verification

- focused backend tests: `16 passed`
- Ruff on all changed Python files: passed
- `git diff --check`: passed
- full backend baseline comparison: `1422 passed, 6 failed`; the same six Windows/path/prompt assertion failures reproduce on clean `origin/main` and are unrelated to this change
- sensitive-data scan: no credentials, tokens, environment files, PDFs, archives, databases, or logs included

Closes #509

### DIFF
--- a/src/backend/app/api/admin_settings.py
+++ b/src/backend/app/api/admin_settings.py
@@ -213,7 +213,7 @@ async def set_literature_search_settings(
 ) -> LiteratureSearchSettings:
     try:
         saved = await literature_settings_service.update_settings(
-            session, payload.model_dump(exclude_none=True)
+            session, payload.model_dump(exclude_unset=True)
         )
     except literature_settings_service.InvalidLiteratureSettingError as exc:
         raise HTTPException(
@@ -234,9 +234,11 @@ async def test_literature_provider(
     source = payload.source.strip().lower()
     started = time.perf_counter()
     try:
-        from app.services.literature.runtime import _default_registry
+        from app.services.literature.runtime import build_adapter_registry
 
-        registry = await _default_registry()
+        registry = await build_adapter_registry(
+            await literature_settings_service.get_runtime_settings(session)
+        )
         adapter = registry.get(source)
         if adapter is None:
             raise ValueError(f"unsupported source: {source}")
--- a/src/backend/app/api/admin_settings.py
+++ b/src/backend/app/api/admin_settings.py
@@ -34,6 +34,7 @@ from app.services import lab as lab_service
 from app.services import literature_settings as literature_settings_service
 from app.services import managed_command_watchdog as watchdog_service
 from app.services import tts as tts_service
+from app.services.literature.multi_source import provider_failure_detail
 
 router = APIRouter(
     prefix="/admin/settings", tags=["admin-settings"], dependencies=[Depends(require_admin)]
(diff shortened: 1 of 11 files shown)


yyy-OPS:
Follow-up hardening is now included in `c1f56c0`.

- redacts provider URLs, query strings, and headers from probe/runtime errors
- stores stable failure codes such as `HTTP_401`, `TIMEOUT`, and `NETWORK_ERROR`
- serializes credential-pool read/modify/write operations, including the first PostgreSQL singleton-row write via a transaction advisory lock

Verification after the update: Ruff passed; administrator credential/runtime suite: `18 passed, 2 warnings`.

tricktreat:
Landed as #526 (merged), rebased onto current main with authorship preserved.

Worth recording one resolution. My first pass took this branch's `runtime.py` and reds seven tests — that copy predates #525, so it silently dropped the progress reporting (`progress["deduplicated"]` and friends). Redone as main's version with only this PR's own change applied on top, which is three lines:

```diff
-    values = configured.get(source) if isinstance(configured, Mapping) else None
+    declared = isinstance(configured, Mapping) and source in configured
+    values = configured.get(source) if declared else None
-    if pool:
+    if declared:
```

That distinction is the substance of the disabled-pool fix, and it is a good catch: judging only by emptiness makes "configured but emptied" indistinguishable from "never configured", so disabling a provider becomes a no-op and the first thing that tells you is the bill, not an error. Added a test pinning all four cases; reverting `if declared:` to `if pool:` reds it.

Also removed a duplicate `score_weights` key (F601) that the test-file union produced.

PR #519, opened 2026-08-28
feat(literature): enrich candidates with versioned venue metrics

### LINKED ISSUE #510: feat(literature): enrich candidates with versioned venue metrics
## Problem

Discovery results do not have a stable journal or conference metrics contract. Missing venues can become display placeholders, and external metric failures cannot be distinguished from a real zero score.

## Scope

- Add OpenAlex and EasyScholar metric providers behind one adapter contract.
- Cache metric responses with provider, fetched-at and expiry metadata.
- Persist a versioned metric snapshot on discovery hits without mutating source metadata.
- Feed verified metrics into the impact dimension only.
- Keep unresolved venues and unavailable metrics as null values.
- Isolate provider timeouts and rate limits from the discovery run.

## Acceptance criteria

The same venue identity resolves consistently across sources. A metric-provider failure leaves the candidate usable. The system never stores `Unknown journal` as a venue or metric value.

Depends on the retrieval quality contract and provider credential lifecycle.

yyy-OPS:
## Summary

- add a shared OpenAlex/EasyScholar venue-metric provider contract
- normalize venue identity by ISSN-L/ISSN, with a canonical-name fallback
- cache resolved and not-found results with provider, fetched-at, and expiry metadata
- persist versioned metric snapshots on discovery hits without mutating source metadata
- feed verified venue metrics into the impact dimension only
- expose EasyScholar through the existing administrator credential lifecycle and connection test
- keep candidates usable when a metric provider times out, rate-limits, or returns invalid data

## Dependency and merge order

This is a stacked backend PR because the metric contract uses the retrieval-quality runtime and administrator credential lifecycle, and its migration follows the current interdisciplinary scope head.

Required merge order:

1. #504
2. #505
3. #516
4. #517
5. this PR

After the dependencies merge, GitHub will reduce the repeated ancestor diff by content. The functional commit for this issue is `56a2168`.

## Persistence and rollback

- adds `literature_search_hits.venue_metric_snapshot`
- adds `literature_venue_metric_cache`
- downgrade removes only the metric snapshot column and cache table
- unresolved venues and unavailable metrics remain null; `Unknown journal` is never persisted as a venue metric

## Tests

- focused venue metric, discovery runtime, administrator settings, ranking, and migration suite: `42 passed`
- expanded literature backend suite: `88 passed`; one unrelated baseline failure remains on Windows because Alembic reads the UTF-8 `alembic.ini` with the GBK locale in `test_migrations_wiki_unify.py`
- affected-file Ruff checks pass
- migration upgrade/downgrade passes
- staged-file sensitive-data scan is clean; no `.env`, key, PDF, database, log, or lockfile is included

Closes #510


tricktreat:
Landed as #527 (merged), rebased onto current main with authorship preserved.

The migration side is clean — `f2a3b4c5d6e7` chains correctly and the duplicate `f1a2b3c4d5e6` it carries is byte-identical to the merged one. Real improvement over the earlier series.

But the branch predates #524 and #526, so two things would have been silently reverted:

- `api/interdisciplinary.py` restores `project.id` in the `IntegrityError` recovery path, undoing the `MissingGreenlet` fix from #524, and removes the `_existing_dedicated_library` helper that makes the race testable at all. Main's version kept.
- `tests/test_literature_discovery_runtime.py` drops both credential guards (from #523 and #526). Union-resolved.

This is the recurring shape: a branch cut before a fix lands carries the pre-fix file, and merging it un-fixes things with nothing failing. Rebasing onto current main right before pushing is what stops it — the migration chain here shows you are already doing that for schema; the same habit applied to the whole branch would have caught these two.

`runtime.py` resolved hunk by hunk rather than by side: your venue-metric wiring taken, #526's `_credential_pool` distinction kept.

PR #520, opened 2026-08-28
feat(literature): schedule incremental library discovery runs

### LINKED ISSUE #511: feat(literature): schedule incremental library discovery runs
## Problem

The existing incremental ingest path reads the daily paper pool. It does not schedule the new multi-source discovery runtime for a specific library or persist those results in its screening pool.

## Scope

- Persist at most one enabled discovery schedule per library.
- Store local execution time, IANA time zone, requested count, year range and settings version.
- Reuse the existing ARQ and task infrastructure.
- Create ordinary discovery runs whose results remain candidates until a curator promotes them.
- Deduplicate against prior discovery hits and formal library papers.
- Enforce creator or curator write access and public-library read-only access.

## Acceptance criteria

Scheduled and manual runs use the same runtime and result contract. A retry cannot create duplicate runs for one schedule window. Disabling a schedule prevents future enqueue operations without deleting history.

Depends on the retrieval quality contract.

yyy-OPS:
## Summary

- persist one versioned incremental-discovery schedule per direction library
- store local execution time, IANA time zone, requested count, candidate budget, year window, next run, and dispatch health
- refactor manual and scheduled run creation through one service contract so library keywords, exclusions, rubric, administrator defaults, sources, and interdisciplinary profiles cannot diverge
- dispatch due schedules through the existing ARQ worker every five minutes
- use row locks, a partial unique index, deterministic job IDs, and queued-run recovery to prevent duplicate active scheduled runs
- keep scheduled results in the ordinary candidate screening pool
- remove candidates already seen in prior discovery history or already present in the formal library, with bounded query batches for a 1,000-candidate pool
- enforce the existing administrator/creator/curator write policy while keeping public schedules read-only for other users

## Dependency and merge order

This is a stacked backend PR. It follows the retrieval quality, credential, interdisciplinary profile, and venue metric contracts, and its migration follows the venue-metric head.

Required merge order:

1. #504
2. #505
3. #516
4. #517
5. #519
6. this PR

After its ancestors merge, GitHub will reduce repeated ancestor changes by content. The functional commit for this issue is `b5bb8de`.

## Persistence and rollback

- adds `literature_discovery_schedules`
- adds `trigger`, `schedule_version`, and `scheduled_for` to `literature_search_runs`
- adds a partial unique index allowing at most one queued/running scheduled run per library
- adds the runtime `tzdata` dependency so IANA time zones behave consistently on Windows and Linux
- downgrade removes only schedule state and run trigger metadata; historical manual discovery data is otherwise unchanged

## Tests

- complete literature/admin/interdisciplinary retrieval/worker/migration suite: `104 passed`
- focused API, scheduler, real ARQ dispatch, runtime incremental filtering, 1,000-candidate batching, migration, and worker registration suite: `40 passed`
- affected backend Ruff checks pass
- Python compileall passes
- migration upgrade/downgrade passes
- sensitive-data scan is clean; no `.env`, key, PDF, ZIP, database, log, or lockfile is included

Closes #511


tricktreat:
Landed as #528 (merged), rebased onto current main with authorship preserved. Migration chains correctly.

Extracting run creation into `discovery_runs.create_discovery_run` is the right move, and I checked it rather than assuming — that kind of refactor is exactly where recently-added behaviour gets left behind. Everything added since #523 came across: administrator defaults, `normalized_score_weights`, `library_discovery_config`, `apply_profile_to_query_plan`, and the `provider_keys` stripping. Behaviourally confirmed too: removing that `pop` from its **new** location still reds `test_provider_keys_never_reach_the_run_snapshot`, so the guard followed the code rather than being stranded at the old call site.

`api/interdisciplinary.py` again predates #524 and restores `project.id` in the recovery path — third PR in a row carrying that file. Main's version kept, both credential guards preserved.

PR #521, opened 2026-08-28
feat(literature): persist translated discovery metadata

### LINKED ISSUE #512: feat(literature): persist translated discovery metadata
## Problem

Discovery candidates have no translation API or persisted translation state. A frontend cannot show reliable progress, retry failures or reuse an existing translation.

## Scope

- Add single-hit and bounded batch translation endpoints.
- Translate title, abstract and inclusion rationale through a dedicated `translation` LLM route.
- Cache by source-text hash, target language and model version.
- Persist queued, running, ready and failed states with stable error codes.
- Keep original metadata unchanged and return translations as an optional view.

## Acceptance criteria

Repeated requests are idempotent, model failures remain retryable, and one failed item does not fail a batch. This issue does not include full-PDF translation.

Depends on #504 and the unified score contract.

yyy-OPS:
## Summary

- add single-hit and bounded batch translation endpoints for literature discovery candidates
- persist title, abstract, and inclusion-rationale translations without mutating source metadata
- cache translations by source hash, normalized target language, and resolved model version
- expose queued, running, ready, and failed lifecycle states with stable retryable error codes
- run translation through a dedicated `translation` LLM stage and registered ARQ worker
- keep public libraries read-only: only discovery managers may create translation jobs

## API

- `POST /api/libraries/{library_id}/literature/runs/{run_id}/hits/{hit_id}/translation`
- `POST /api/libraries/{library_id}/literature/runs/{run_id}/translations`
- `GET /api/libraries/{library_id}/literature/runs/{run_id}/hits/{hit_id}/translation`

The batch endpoint accepts at most 100 hit IDs. A queue-dispatch failure is recorded on that item and does not abort the rest of the batch.

## Persistence

Migration `f4a5b6c7d8e9` adds `literature_hit_translations` after the incremental-discovery migration. The unique cache identity is:

`hit_id + target_language + source_hash + model_version`

Original discovery-hit fields remain unchanged. Failed rows can be requested again and transition back to `queued`.

## Stack and review order

This is a stacked PR based on #520 (`b5bb8de`) and targets `main`, consistent with the preceding literature PR series. Review the new commit `1b81ec7` for this issue. Recommended merge order for this chain:

`#504 -> #505 -> #516 -> #517 -> #519 -> #520 -> this PR`

#507 is independent from this translation cache and can be reviewed separately.

## Verification

- `133 passed` across literature discovery, scheduling, OA cache, retrieval quality, venue metrics, LLM routing, worker registration, and migration tests
- `3 passed` for the frontend/backend LLM stage registry contract
- Ruff passed for changed backend modules and tests
- `compileall` passed
- Alembic SQLite upgrade/downgrade round trip passed

Closes #512


yyy-OPS:
Follow-up correctness fix in `1b3201e`:

- persist the user who requested a translation job
- resolve the worker's translation route with that user instead of the discovery-run creator
- keep scheduled runs and curator-initiated translations from failing with `TRANSLATION_MODEL_CHANGED`

Fresh verification after the fix:

- combined literature/admin/interdisciplinary/worker/migration suite: `115 passed` 
- focused translation/worker/migration suite: `15 passed` 
- affected-file Ruff and `git diff --check`: passed

tricktreat:
Landed as #529 (merged), rebased onto current main with authorship preserved. Migration `f4a5b6c7d8e9` chains correctly.

`api/interdisciplinary.py` predates #524 for the fourth consecutive PR in this batch and would again have reverted the `MissingGreenlet` fix. Main's version kept; runtime tests union-resolved so both credential guards survive.

Worth flagging the pattern rather than just the instance: none of those four PRs touch interdisciplinary confirmation as a feature — the file rides along because the branches were cut from a common point and never rebased afterwards. Your migration chain in this same batch is clean and correctly linked, so the schema half of rebasing is clearly being done; extending the same step to the whole branch immediately before pushing removes this entire class of conflict.

One from my side: `app/schemas/literature_discovery.py` and `app/models/literature_discovery.py` share a basename and my first resolution pass only handled the models one, leaving conflict markers in the schemas file. `import app.main` caught it before anything was pushed.

PR #522, opened 2026-08-28
feat(reading): serve authorized structured paper content

### LINKED ISSUE #513: feat(reading): serve authorized structured paper content
## Problem

Content versions persist Markdown paths, text, chunks and MinerU assets, but the API does not deliver a complete authorized structured-content manifest. The reader would have to inspect storage paths or guess asset URLs.

## Scope

- Return the current content-version status and manifest for a library paper.
- Serve UTF-8 Markdown, figures and tables through authorized endpoints.
- Use short-lived signed asset URLs, stable content hashes and cache headers.
- Expose parser, page count, parse status and vector states without storage paths.
- Preserve library visibility and active AssetGrant checks for every resource.

## Acceptance criteria

Private and institution-authorized assets cannot be read through another library. Markdown references resolve to the persisted version assets. PyMuPDF content remains available as plain structured text without claiming MinerU figures or tables.

Depends on #507.

yyy-OPS:
## Summary

- return the current parsed-content status and a storage-path-free manifest for a library paper
- serve UTF-8 MinerU Markdown, figures, tables, and plain parsed text through authorized endpoints
- rewrite persisted relative Markdown/HTML asset references to short-lived signed URLs
- expose parser, parser version, page/chunk counts, parse status, and both vector states
- return stable SHA-256 content/asset hashes and bounded private cache headers
- keep PyMuPDF fallback explicit as `plain_text` without claiming MinerU figures or tables

## Authorization and storage safety

Every manifest request verifies library visibility, paper membership, and an active readable `AssetGrant`. Every signed resource request repeats those checks, so revocation takes effect before URL expiry. Signed URLs are bound to the requesting user, library, paper, content version, relative path, and expiry.

Persisted storage paths are never returned. Resource resolution rejects absolute paths, traversal, unsafe manifest entries, forged tokens, and expired tokens. HTML/SVG/table responses use a sandboxed content-security policy.

## API

- `GET /api/libraries/{library_id}/papers/{paper_id}/structured-content`
- `GET /api/structured-content-assets/{signed_token}`

`POLARIS_STRUCTURED_CONTENT_LINK_TTL_SECONDS` defaults to 300 seconds and is clamped to 60–3600 seconds.

## Dependency and review order

This PR starts directly from `origin/main@192f019` and contains three commits:

1. `22d20e5` — patch-equivalent to #507 commit `3bb45d9`
2. `eb9544b` — patch-equivalent to #507 commit `670272b`
3. `da42af3` — the #513 structured-content API

Recommended order: merge #507 first, then review the remaining `da42af3` delta here. Git will remove the patch-equivalent #507 changes from this PR after #507 lands. This PR does not depend on the literature discovery/translation stack.

## Verification

- `37 passed` across structured-content API, MinerU lifecycle, content versions, PDF assets, evidence anchors, AI full-text context, and download-batch protocol
- Ruff passed for the backend application, worker, and new tests
- `compileall` passed

A full Windows run produced `1473 passed, 6 failed`. The same six failures reproduce unchanged on `origin/main@192f019`: three locale/path failures (CP936 Alembic/API reads and slash normalization) and three existing experiment prompt/state assertions. None touches this PR's files or its verified paths.

Closes #513

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -24,6 +24,8 @@ POLARIS_INVITE_CODE=polaris-lab
 # POLARIS_PUBLIC_BASE_URL=https://polaris.example.edu
 # Short-lived paper-figure download URL lifetime, in seconds.
 POLARIS_MCP_DOWNLOAD_LINK_TTL_SECONDS=900
+# Short-lived structured Markdown, figure, and table URL lifetime, in seconds.
+POLARIS_STRUCTURED_CONTENT_LINK_TTL_SECONDS=300
 
 # ---- Database / Cache ----
 POLARIS_DATABASE_URL=postgresql+asyncpg://polaris:polaris@postgres:5432/polaris
--- a/src/backend/app/api/router.py
+++ b/src/backend/app/api/router.py
@@ -42,6 +42,7 @@ from app.api import (
     shelf,
     skills,
     ssh_credentials,
+    structured_content,
     tts,
     users_profile,
     views,
@@ -72,6 +73,7 @@ api_router.include_router(papers.router)
 api_router.include_router(library.router)
 api_router.include_router(literature_discovery.router)
 api_router.include_router(paper_assets.router)
+api_router.include_router(structured_content.router)
 api_router.include_router(libraries.router)
 api_router.include_router(publications.router)
 api_router.include_router(notes.router)
--- a/src/backend/app/api/structured_content.py
+++ b/src/backend/app/api/structured_content.py
@@ -1,0 +1,289 @@
+"""Authorized manifest and signed resources for parsed paper content."""
+
+import hashlib
+import time
+import uuid
+from datetime import UTC, datetime
+
+from fastapi import APIRouter, Depends, HTTPException, status
+from fastapi.responses import FileResponse, Response
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from app.api.auth import current_active_user
+from app.core.db import get_session
+from app.models.paper_content import PaperContentVersion
+from app.models.user import User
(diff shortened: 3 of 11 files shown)


tricktreat:
Merged as-is — no rebase, no conflicts, nothing to fix.

I verified the security claims rather than reading them, since this PR hands out signed URLs:

**Path handling.** `_safe_relative_path` rejects every traversal form I threw at it — `../x`, `a/../../x`, `/etc/x`, `..\\..\\x`, empty — and normalises `.//x` to `x`.

**Token integrity.** Forged and malformed tokens (``, `abc.def`, a plausible-looking `base64.hex`, `..`, `a.b.c`) are all rejected with `InvalidStructuredContentToken`. Verification uses `hmac.compare_digest`, so it is constant-time, and the HMAC is domain-separated by a signing-context prefix.

**The tests actually bite.** Replacing the `compare_digest` check with `if False:` reds `test_structured_content_tokens_reject_tampering_and_expiry`. A signature check with no test behind it is one refactor away from being decorative; this one is not.

The design decision I want to note approvingly: re-running visibility, membership and `AssetGrant` checks on every signed-resource request rather than trusting the token alone. That is what makes revocation take effect before expiry, and `test_cross_library_access_is_denied_and_revocation_invalidates_signed_url` covers it. Keeping persisted storage paths out of the manifest entirely is the right call too — a path that is never returned cannot be probed.

26 passed across `test_paper_reading`, `test_paper_content`, `test_paper_assets`, `test_migrations`; `ruff check app/ worker/` clean; single alembic head `f4a5b6c7d8e9`.

Issue #515, opened 2026-08-27
feat(interdisciplinary): carry profile versions across project workflows


yyy-OPS:
## Problem

Interdisciplinary profiles, dedicated libraries and retrieval matrices are persisted, but later project workflows do not consistently record or apply the confirmed profile and Skill template versions.

## Scope

- Bind project runs to one confirmed profile version and one built-in Skill template version.
- Carry primary discipline, associated disciplines, bridge questions, terms and evidence coverage into ideas, experiments, writing and review.
- Persist the versions used by each run for reproducibility.
- Keep conventional projects on their existing prompts and workflow.
- Reject placeholder discipline values before a profile is confirmed.

## Acceptance criteria

A confirmed interdisciplinary project uses the same versioned constraints across downstream runs. Updating the profile creates a new version without rewriting prior run snapshots. Conventional project regression tests remain unchanged.

Depends on #505 and the interdisciplinary work merged in #502 and #503.

yyy-OPS:
Reopened because #525 closed this issue accidentally. #525 implements retrieval-quality issue #508, not interdisciplinary profile-version propagation. Carrying confirmed profile and skill versions through idea, experiment, writing, review, and other project workflows remains to be implemented.

PR #530, opened 2026-08-28
feat(interdisciplinary): pin workflow profile versions

yyy-OPS:
## Summary

- freeze the latest confirmed interdisciplinary profile version into each project Voyage run
- pin the current built-in interdisciplinary workflow Skill version in the same immutable checkpoint
- inject the confirmed scope, disciplines, bridge questions, evidence boundary, validation conditions, query channels, and evidence balance into Wiki, idea, experiment, writing, review, and presentation prompts
- reject placeholder discipline names before profile confirmation
- preserve the existing behavior for conventional projects

## Design

This PR reuses `VoyageRun.checkpoint` instead of adding schema fields or a migration. The first drive snapshots both version IDs and the complete runtime context. Resume and replay read that immutable snapshot, while later runs can bind newer profile or Skill versions.

The built-in workflow remains visible through the existing Skill system. Interdisciplinary guidance is adapted to existing prompt targets, so downstream agents consume one shared contract rather than adding parallel workflow implementations.

## Verification

- `28 passed` across interdisciplinary profiles, retrieval, built-in Skills, Skill execution, version pinning, and conventional-project isolation
- `150 passed` across Voyage, idea generation/review, experiments, writing, paper review, and presentations
- Ruff passed for all changed backend files
- `compileall` passed

The only warnings are existing FastAPI/Starlette deprecation warnings for `HTTP_422_UNPROCESSABLE_ENTITY` in unchanged routes.

## Review and rollback

- Based directly on `origin/main@851abc8` after #522 merged
- One commit; no dependency branch or duplicated changes from earlier PRs
- No migration and no persisted-data rewrite
- Rollback is limited to removing the prompt adapter/service and its three call sites

Refs #515. The issue is currently closed because of an earlier incorrect auto-close reference; this PR implements its remaining acceptance criteria.

### DIFF
--- a/src/backend/app/agents/voyage/actions_experiment.py
+++ b/src/backend/app/agents/voyage/actions_experiment.py
@@ -497,6 +497,7 @@ def _prompt_with_context(base: str, ctx: ActionContext) -> str:
     """按 params.eval_model / hf_mirror / extra_notes 给 system prompt 条件追加段落。"""
     params = _params(ctx)
     parts = [base, STACK_GUARD_PROMPT_SECTION]
+    parts.append(ctx.skill_guidance("experiment.plan"))
     if str(params.get("eval_model") or "").strip():
         parts.append(EVAL_MODEL_PROMPT_SECTION)
     if params.get("hf_mirror"):
--- a/src/backend/app/agents/voyage/engine.py
+++ b/src/backend/app/agents/voyage/engine.py
@@ -46,6 +46,7 @@ from app.models.base import utcnow
 from app.models.gate import Gate
 from app.models.llm_config import LLMUsage
 from app.models.voyage import TERMINAL_STATUSES, VoyageRun, VoyageStep, mode_for_kind
+from app.services import interdisciplinary_workflows
 from app.services import skills as skills_service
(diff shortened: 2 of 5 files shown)


tricktreat:
Landed as #548 (merged) with authorship preserved, plus one test.

Reusing `VoyageRun.checkpoint` instead of adding schema is the right call for something per-run and immutable, and saying so in the PR rather than leaving it to be inferred is appreciated.

The immutability claim holds — `_ensure_skills_snapshot` returns early once `"skills"` is present. But the existing test covers only the other direction (a **new** run picks up v2); nothing pinned that an **in-flight** run keeps v1, which is the headline guarantee. Added it, because that is where the failure would hurt: a run whose context changes mid-flight produces its first half under one scope and its second half under another, and the replay shows only the later one — artefacts and record disagree, with no error anywhere. Removing the early-return guard reds the new test.

One edge I did not change: the guard keys on `"skills"`, so a voyage already in flight when this deploys has `skills` but no `interdisciplinary_context` and will never receive one. Same as today's behaviour and it self-corrects next run — but if you would rather in-flight runs pick it up on resume, the guard needs to key on both.

Issue #514, opened 2026-08-27
feat(ai): apply versioned full-text evidence across research workflows


yyy-OPS:
## Problem

Library chat and MCP can use versioned full text, but Wiki compilation, ideas, experiments, writing, review and presentations do not share one evidence-context contract.

## Scope

- Extract one authorized evidence context builder for AI workflows.
- Record the content version, chunks and evidence manifest actually sent to the model.
- Emit sentence-level anchors only for supplied and resolved text.
- Mark abstract-only inputs explicitly.
- Fall back to a paper-level citation when sentence resolution fails.
- Apply the contract to Wiki, ideas, experiments, writing, review and presentation generation without changing their domain outputs.

## Acceptance criteria

Every fine-grained citation resolves to a supplied full-text chunk. Reparse keeps old references usable through quote remapping or paper-level fallback. Unauthorized full text never enters a prompt, cache or generated artifact.

Depends on the structured content API and the evidence work already merged in #500 and #501.

PR #531, opened 2026-08-28
feat(ai): ground research workflows in versioned full text

yyy-OPS:
## Summary

- build one immutable, authorized evidence snapshot for project AI workflows
- carry content-version, chunk, sentence-anchor, parser, page, and reader locator metadata
- inject the same evidence contract into Wiki compilation, ideas, experiments, writing, paper review, and presentations
- persist only sentence references that were both supplied to and emitted by an action
- explicitly mark full-text, mixed, abstract-only, and unavailable evidence modes
- preserve reparse navigation through the existing anchor resolver and its quote/chunk/paper fallback

## Authorization and artifact safety

Project evidence is read only through source-library `AssetGrant` checks. Revoking a grant immediately excludes the parsed content. PDF text is treated as untrusted source material and cannot override system or user instructions.

Voyage checkpoints are project-scoped and record the exact model-visible evidence snapshot. The globally shared per-paper Wiki stores only version/chunk/anchor locators in its hidden manifest; it never copies authorized PDF sentences into that artifact. Resolving a locator still passes the existing library and readable-asset authorization checks.

## Performance

Sequential full-text pagination now uses database `OFFSET/LIMIT + 1` instead of loading every chunk for every page. Query-ranked retrieval keeps the existing in-memory ordering behavior. Evidence snapshots are built once per target Voyage run and reused by all subsequent steps.

## Dependency and merge order

- Based directly on `origin/main@851abc8`
- Depends on the structured-content and evidence foundation already merged through #522, #500, and #501
- Does not depend on #530
- Git `merge-tree` checks passed in both directions with #530 and produced the same conflict-free merge tree
- One commit; no migration or duplicated dependency commits

## Verification

- `63 passed` across evidence snapshots, authorization revocation, reparse fallback, structured content, Wiki compilation, figure handling, and affiliation extraction
- `150 passed` across Voyage, ideas, experiments, writing, paper review, and presentations
- Ruff and `compileall` passed

Warnings are limited to existing SQLite/Alembic and FastAPI/Starlette deprecation warnings in unchanged code.

Refs #514. The issue is currently closed due to an earlier incorrect auto-close reference; this PR implements its remaining acceptance criteria.

### DIFF
--- a/src/backend/app/agents/voyage/actions.py
+++ b/src/backend/app/agents/voyage/actions.py
@@ -16,6 +16,7 @@ from app.core.events import EventBus
 from app.core.llm.base import Message
 from app.core.llm.router import LLMRouter
 from app.models.voyage import VoyageRun
+from app.services.ai_evidence_context import evidence_guidance
 
 
 @dataclass(slots=True)
@@ -58,6 +59,10 @@ class ActionContext:
     def skill_personas(self, target: str) -> list[dict[str, Any]] | None:
         """persona 技能人设列表；None = 用调用方内置默认。"""
         return skill_personas(self.checkpoint, target)
+
+    def evidence_guidance(self) -> str:
+        """返回本次 Voyage 固定的授权全文证据；未配置时为空。"""
+        return evidence_guidance(self.checkpoint.get("ai_evidence"))
 
 
 ActionFunc = Callable[[ActionContext, dict[str, Any]], Awaitable[dict[str, Any]]]
--- a/src/backend/app/agents/voyage/actions_experiment.py
+++ b/src/backend/app/agents/voyage/actions_experiment.py
@@ -514,7 +514,7 @@ def _prompt_with_context(base: str, ctx: ActionContext) -> str:
         ]
         if qa_lines:
             parts.append(INTAKE_PROMPT_SECTION.format(qa="\n".join(qa_lines)))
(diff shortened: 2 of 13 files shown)


tricktreat:
Landed as #549 (merged) with authorship preserved, plus one test.

The claim carrying real risk here is that authorized full text never reaches the globally shared per-paper Wiki — a sentence landing there is readable without the `AssetGrant`, persistently, and revoking afterwards does not take it back. It holds, and for the right reason: `append_evidence_manifest` builds the persisted refs from an explicit **allowlist** rather than stripping known-bad keys. That fails closed when the bundle grows a field; a denylist would not. Added a test pinning it — a manifest entry carrying `quote`/`quoted_text`/`text` yields an artifact with none of them, while the anchor id survives so reader navigation still resolves.

One thing that is prompt-level rather than enforced, for you to decide on: the manifest is code-guaranteed, but the Wiki **body** is model output and the model is shown the sentences. Nothing prevents it reproducing them there, and the existing test only checks the region after the manifest marker. I have not changed it — scrubbing evidence sentences from generated prose could damage legitimate output. But the guidance wording "Cite the exact supplied sentence as [文N·句M]" reads slightly toward including the sentence; "cite by locator, do not reproduce the sentence text" would lean the safe way at no cost.

PR #533, opened 2026-08-28
feat(literature): add library discovery workspace

### LINKED ISSUE #532: feat(literature): add library discovery workspace UI
## Problem

The multi-source discovery backend is available, but a library curator still has no first-class workspace in the Polaris literature library UI. Users cannot start a run, inspect source progress, review abstracts and score dimensions, translate metadata, filter history, or promote selected candidates without calling APIs directly.

## Scope

- Add **Discover literature** as the leftmost tab inside a concrete literature library.
- Show library-scoped run creation, live phase/source progress, bounded history, and permanent history deletion.
- Render candidate title, abstract, authors, venue, DOI, OA/PDF state, source evidence, score dimensions, result lane, and selection rationale.
- Add search, year/PDF/discovery filters, discipline/evidence-role filters, and relevance/novelty/impact sorting.
- Support metadata translation, candidate selection, promotion into the current library, and download-batch dispatch through existing APIs.
- Enforce curator write controls and public-library read-only behavior in the interface.
- Keep the page visually and structurally aligned with the existing library workbench.

## Non-goals

- Administrator provider-key settings.
- Structured-content reader and evidence navigation.
- Browser extension implementation.
- New discovery backend contracts or database migrations.

## Acceptance criteria

A curator can run and review multi-source discovery entirely inside a library, while a read-only visitor can inspect persisted results without seeing mutation controls. Progress identifies the active phase and source attempts. The frontend build, focused interaction tests, and existing library-workbench regression tests pass.

Depends only on the discovery, ranking, translation, metrics, schedule, OA promotion, and download-batch APIs already merged into `main`.

yyy-OPS:
## Summary

- add **Discover** as the leftmost tab inside each managed or public library workspace
- render persisted discovery runs with live phase and per-source progress, paginated candidates, search/year/source/status filters, ranking dimensions, venue metrics, rationale, DOI/OA state, and a compact history modal
- expose translation, OA cache, candidate promotion, cancellation, and permanent history deletion with explicit loading and error feedback
- create one persistent Polaris extension batch for selected, already-promoted papers while preserving independent library/paper bindings and skipping server-cached PDFs
- add server-side publication-year filters so result totals and pagination remain correct

## Authorization and lifecycle

- reuses the existing can_manage contract (platform admins, library creators, and curators); public-library readers receive a strictly read-only discovery view
- discovery hits remain a review pool: no Paper/PDF asset parsing or vectorization is triggered until promotion into the library
- no database migration is required

## Merge independence

- based directly on main at 851abc8
- does not depend on #530 or #531 and can be merged independently

## Verification

- frontend test suite: 94 passed
- frontend production build: passed
- literature discovery API and extension batch protocol: 6 passed
- Ruff for touched backend files: passed
- Chrome visual checks: 1360px desktop, 390px mobile, history modal, and public read-only view

Closes #532
## Follow-up fix

The backend batch is now the source of truth for extension dispatch. The page sends only `queued` library and paper bindings, keeps cached `skipped` items in the durable batch history, and emits no extension event when every selected paper already has a readable PDF. Binding IDs, rather than list positions, determine which papers are sent.

The updated branch passes 95 frontend tests, the production build, and 12 focused discovery and download protocol tests.

### DIFF
--- a/src/backend/app/api/literature_discovery.py
+++ b/src/backend/app/api/literature_discovery.py
@@ -274,12 +274,19 @@ async def list_hits(
     hit_status: str | None = Query(
         None, alias="status", pattern="^(candidate|promoted|dismissed)$"
     ),
+    year_from: int | None = Query(None, ge=1800, le=3000),
+    year_to: int | None = Query(None, ge=1800, le=3000),
     sort: str = Query("relevance", pattern="^(relevance|novelty|impact|recent|title)$"),
     page: int = Query(1, ge=1),
     size: int = Query(20, ge=1, le=100),
     session: AsyncSession = Depends(get_session),
     user: User = Depends(current_active_user),
 ) -> SearchHitPage:
+    if year_from is not None and year_to is not None and year_from > year_to:
+        raise HTTPException(
+            status.HTTP_422_UNPROCESSABLE_ENTITY,
+            detail="YEAR_RANGE_INVALID",
+        )
     await _library(session, library_id, user)
     run = await discovery_runs.get_visible_run(session, library_id=library_id, run_id=run_id)
     if run is None:
@@ -289,6 +296,10 @@ async def list_hits(
         stmt = stmt.where(LiteratureSearchHit.source == source.strip().lower())
     if hit_status:
         stmt = stmt.where(LiteratureSearchHit.status == hit_status)
+    if year_from is not None:
+        stmt = stmt.where(LiteratureSearchHit.year >= year_from)
+    if year_to is not None:
+        stmt = stmt.where(LiteratureSearchHit.year <= year_to)
     if q:
         pattern = f"%{q.strip()}%"
         stmt = stmt.where(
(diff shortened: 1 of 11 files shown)


tricktreat:
Merged as-is. Based on current main, no migration, clean merge; `tsc --noEmit` and `vitest run` (95 tests) both pass, backend 28 passed, ruff clean.

Two things I checked:

**The year-filter claim holds, and for a reason worth knowing.** `year_from`/`year_to` are applied to `stmt` before it is executed, and `list_hits` computes `total = len(hits)` from that same materialised result — so totals and pagination cannot drift from the filters. Note that in-memory pagination is pre-existing on main, not something this PR introduced; it does mean every page request materialises the whole result set and sorts it in Python. Fine at review-pool sizes, worth revisiting if `candidate_budget` grows.

**i18n follows the project convention exactly.** Module-level constants hold `['中文', 'English']` tuples and `tr()` is called at render time (84 sites), rather than calling `tr()` at module level — which is the failure mode this codebase has hit before.

Minor, non-blocking: the new 422 uses `HTTP_422_UNPROCESSABLE_ENTITY`, which Starlette now deprecates in favour of `HTTP_422_UNPROCESSABLE_CONTENT`. The repo currently has 49 of the former and 12 of the latter, so this matches the majority — just noting it since the warning shows up in test output.

PR #526, opened 2026-08-28
feat(admin): add literature provider credential pool lifecycle

tricktreat:
#517 merged, rebased onto current `main`. Authorship preserved as @yyy-OPS.

Its branch is based on #504's head, which landed as #523, and #516 landed as #525 in between — so `runtime.py` and `api/literature_discovery.py` conflicted with the retrieval work.

## The resolution that mattered

My first pass took this PR's side for `runtime.py` and reds seven tests: that copy predates #525, so it silently dropped the progress reporting (`progress["deduplicated"]` and friends). Redone as `main`'s version with only this PR's own change applied on top, which turns out to be three lines:

```diff
-    values = configured.get(source) if isinstance(configured, Mapping) else None
+    declared = isinstance(configured, Mapping) and source in configured
+    values = configured.get(source) if declared else None
-    if pool:
+    if declared:
```

That distinction is the substance of *"prevent disabled credential pools from silently falling back to environment credentials"*. Judging only by whether the pool is empty makes "configured but emptied" indistinguishable from "never configured", so disabling a provider becomes a no-op — requests keep going out on the environment credentials, and the first thing that tells you is the bill or the rate limit, not an error.

For `api/literature_discovery.py` the three conflicting hunks were the reverse: this branch carries the pre-#525 state, so `main`'s side was kept.

## Added

A test pinning the four cases of `_credential_pool` — unconfigured falls back, `{}` falls back, `[]` and `["  "]` do not, populated wins. Reverting `if declared:` to `if pool:` reds it.

Also removed a duplicate `score_weights` dict key that the test-file union produced (`F601`); the surviving one is #525's five-dimension normalized form.

## Verification

- `alembic heads` → single head `f1a2b3c4d5e6` (no migration in this PR)
- `tests/test_admin_literature_settings.py test_literature_discovery_runtime.py test_literature_discovery_api.py test_migrations.py` → 31 passed
- `ruff check app/ worker/` and the touched test files → clean

Closes #518. Supersedes #517.

PR #535, opened 2026-08-28
feat(admin): add literature search settings workspace

### LINKED ISSUE #534: feat(admin): add literature search settings workspace
## Context

The backend already exposes administrator-scoped literature discovery settings and credential-pool lifecycle endpoints through #526, but the administrator console has no UI for them. Operators currently cannot inspect enabled sources, tune retrieval defaults, rotate provider keys, or run connection probes without calling the API manually.

## Scope

- Add a `Literature search` tab to the existing administrator settings page.
- Present retrieval defaults, enabled sources, scoring weights, provider health, and credential pools in clear sections that follow the current Polaris settings design system.
- Add modal-based credential create/edit flows. Never return or prefill stored secrets; editing a credential may optionally replace its secret.
- Support per-provider and per-key connection tests with visible pending/success/failure feedback.
- Make keyless providers such as arXiv explicit and support multiple keys for providers such as OpenAlex.
- Keep all changes frontend-only and use the API contracts already merged in #526.

## Acceptance criteria

- Administrators can read and update discovery defaults and source selection.
- Administrators can add, edit, enable/disable, test, and delete individual provider credentials.
- Secret values are never exposed after submission.
- Provider health and last test results are visible.
- The page remains usable on desktop and mobile and matches existing Polaris settings patterns.
- Frontend tests and production build pass.

## Non-goals

- Changing provider adapters, search ranking, database models, or backend authorization.
- Adding a standalone literature-discovery product area.

yyy-OPS:
## Summary

- add a `Literature search` tab to the existing administrator settings console
- expose retrieval defaults, enabled sources, Polaris scoring weights, provider health, and credential pools through the API merged in #526
- add modal-based credential create/edit flows plus per-provider and per-key connection tests
- keep stored secrets write-only: the UI displays only the backend mask and never pre-fills an existing secret
- support multiple rotating keys, explicit keyless sources such as arXiv, and legacy credential pools returned by the backend

## Architecture and merge boundary

This is a frontend-only PR based directly on `main@851abc8`. It depends only on the backend contract already merged in #526 and does **not** depend on #530, #531, or #533. It can be reviewed and merged independently.

The page does not add a standalone literature product area. It stays inside the current administrator settings shell and uses the existing Polaris cards, segmented navigation, switches, modal, confirmation dialog, toast, and responsive layout primitives.

## Security behavior

- credential secrets are accepted only in create/edit modals and are never read back
- editing with an empty secret preserves the current encrypted value
- enable/disable, test, and delete operate on one credential ID at a time
- settings updates omit `provider_keys` and `provider_health`, so saving retrieval defaults cannot replace or erase the credential pool

## Validation

- `npm test`: 17 files, 96 tests passed
- `npm run build`: TypeScript and Vite production build passed
- focused settings contract tests: 5 passed
- desktop (1360px), mobile (390px), credential-pool, and credential-modal visual checks completed
- `git diff --check` passed
- sensitive-value scan passed

Closes #534

### DIFF
--- a/src/frontend/src/features/settings/AdminSettingsPage.tsx
+++ b/src/frontend/src/features/settings/AdminSettingsPage.tsx
@@ -5,6 +5,7 @@ import { EmptyState } from '../../components/ui/EmptyState';
 import { PageHead } from '../../components/ui/PageHead';
 import { Segmented } from '../../components/ui/Segmented';
 import { ExperimentSettings } from './ExperimentSettings';
+import { LiteratureSearchSettingsPanel } from './LiteratureSearchSettings';
 import { FeedbackTab } from '../feedback/FeedbackTab';
 import { tr } from '../../lib/i18n';
 import { api, isAdmin } from '../../lib/api';
@@ -15,9 +16,9 @@ import { CodesTab, DailyCategoriesTab, LlmTab, UsageTab, UsersTab } from './Sett
    各标签页组件仍住在 SettingsPage.tsx（与个人设置共用一批内部小组件），这里只负责壳层。
    ============================================================ */
 
-type AdminTab = 'llm' | 'experiment' | 'daily' | 'users' | 'codes' | 'feedback' | 'usage';
+type AdminTab = 'llm' | 'literature' | 'experiment' | 'daily' | 'users' | 'codes' | 'feedback' | 'usage';
 
-const ADMIN_TABS: AdminTab[] = ['llm', 'experiment', 'daily', 'users', 'codes', 'feedback', 'usage'];
+const ADMIN_TABS: AdminTab[] = ['llm', 'literature', 'experiment', 'daily', 'users', 'codes', 'feedback', 'usage'];
 
 export function AdminSettingsPage() {
   const { data: me, isLoading } = useQuery({ queryKey: ['me'], queryFn: () => api.me(), retry: false });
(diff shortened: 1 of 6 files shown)


yyy-OPS:
Follow-up: commit `13897f9` removes the vendor-specific scoring wording from the visible settings UI. The PR description now uses Polaris-native terminology throughout; no external service is introduced.

tricktreat:
Landed as #550 (merged) with authorship preserved, plus one test tightened.

The credential handling is right — previews only, password inputs, and an edit that leaves the field empty omits `secret` from the payload rather than sending `""` over the stored key. `buildLiteratureSettingsUpdate` builds an explicit field list so masked previews and health state cannot be echoed back. The empty-pool copy also distinguishes "cannot use the administrator key pool" from "anonymous quota or environment fallback", which matches the `declared`-versus-empty distinction landed in #526 — UI and runtime telling the same story.

But the test for that did not test it:

```js
const payload = buildLiteratureSettingsUpdate(validDraft());
expect(payload).not.toHaveProperty('provider_keys');
```

`validDraft()` never carries `provider_keys`, so the assertion held vacuously — replacing the builder body with `return { ...draft }` left all five tests green. Rewritten to pollute the draft with a masked preview and health state first, which is the shape a GET actually returns and therefore what a round-trip would echo. Now `{ ...draft }` reds it.

General point worth carrying forward: an assertion that something is *absent* only means something if the input could have supplied it.

PR #543, opened 2026-08-28
feat(settings): manage Polaris extension API keys

### LINKED ISSUE #542: feat(settings): manage Polaris extension API keys
## Problem

Polaris already exposes a personal API key lifecycle for the browser extension through `POST /me/download-api-key`, `DELETE /me/download-api-key`, and the authenticated identity probe at `GET /download-client/me`. The user settings page has no way to create, copy, test, rotate, or revoke that key, so users cannot complete or verify the extension connection flow from the product UI.

## Proposed scope

- Add a personal settings tab for the Polaris extension.
- Show the current Polaris base URL and provide a copy action.
- Create or rotate a personal extension API key through the existing backend endpoint.
- Show plaintext key material only in the response that created it.
- Test the newly created key against the existing extension identity endpoint.
- Explain that rotation invalidates the previous key.
- Require confirmation before revoking the key.
- Keep the UI honest about server state. The backend does not expose key metadata, so the page must not claim whether a key already exists after reload.

## Acceptance criteria

- The frontend API client matches the existing key lifecycle and identity contracts.
- Users can copy the base URL and the newly created key.
- Users can test the newly created key and see clear success or failure feedback.
- Refreshing the page removes plaintext key material from the UI.
- Rotation and revocation have clear loading, success, and failure feedback.
- Revocation cannot happen through a single accidental click.
- The layout works at desktop and mobile widths.
- Frontend tests and the production build pass.

## Security notes

The page must never persist the plaintext key in local storage, session storage, logs, query strings, or application state outside the mounted settings component.

yyy-OPS:
## Summary

Add a personal settings page for connecting the Polaris browser extension with the API key lifecycle already available on `main`.

- expose the server base URL with a copy action
- create or rotate the personal `pol_dl_` API key
- show plaintext key material only in the component state that receives the create response
- test the new key against the extension identity endpoint and report the connected account
- explain that page reload removes the plaintext value and that rotation invalidates the previous key
- require a confirmation dialog before revocation
- report copy, request, connection-test, and revocation failures to the user
- support `/settings?tab=extension` deep links

## Scope and dependency

This PR is based directly on `main@851abc8`. It has one commit and does not depend on another open PR. The backend endpoints used here are already present on `main`:

- `POST /me/download-api-key`
- `DELETE /me/download-api-key`
- `GET /download-client/me`

The backend does not expose key metadata. The page therefore does not claim whether a key already exists after reload.

## Security behavior

The plaintext key stays only in mounted React component state. This PR does not write it to local storage, session storage, query strings, logs, or repository files.

## Verification

- frontend suite: 18 files, 95 tests passed
- focused extension settings tests: 4 passed
- `npm run build`: passed
- backend download protocol regression: 5 passed
- interaction smoke: create once, identity probe once, plaintext removed after reload, no DELETE before confirmation, one DELETE after confirmation
- visual review: 1440x1000 and 390x844, no horizontal overflow; revocation dialog remains inside the mobile viewport
- `git diff --check`: passed
- sensitive information scan: 0 hits

Closes #542

### DIFF
--- a/src/frontend/src/features/settings/ExtensionApiKeySettings.tsx
+++ b/src/frontend/src/features/settings/ExtensionApiKeySettings.tsx
@@ -1,0 +1,219 @@
+import { useState } from 'react';
+import { useMutation } from '@tanstack/react-query';
+import { FormField } from '../../components/ui/FormField';
+import { Icon } from '../../components/ui/Icon';
+import { Modal } from '../../components/ui/Modal';
+import { toast } from '../../components/ui/Toast';
+import { api } from '../../lib/api';
+import { copyText } from '../../lib/clipboard';
+import { portalUrl } from '../../lib/endpoint';
+import { tr } from '../../lib/i18n';
+
+export function ExtensionApiKeySettings() {
+  const [apiKey, setApiKey] = useState<string | null>(null);
+  const [keyPrefix, setKeyPrefix] = useState<string | null>(null);
+  const [revokeOpen, setRevokeOpen] = useState(false);
+  const baseUrl = portalUrl();
+
+  const rotateMutation = useMutation({
+    mutationFn: () => api.rotateDownloadApiKey(),
+    onSuccess: (result) => {
+      setApiKey(result.api_key);
+      setKeyPrefix(result.key_prefix);
+      toast(
+        tr('API Key 已生成，之前的密钥已失效', 'API key created. The previous key is now invalid.'),
+        'ok',
+      );
+    },
+    onError: (error) => {
+      const message = error instanceof Error ? error.message : String(error);
+      toast(`${tr('生成失败', 'Generation failed')}: ${message}`, 'error');
+    },
+  });
+
+  const revokeMutation = useMutation({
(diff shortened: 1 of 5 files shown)


tricktreat:
Merged as-is. `tsc --noEmit` clean, `vitest run` 117 passed, no migration, clean merge.

The key handling is right on the points that matter for a one-time secret:

- the plaintext lives only in component state and is never written to `localStorage`, `sessionStorage` or `indexedDB` — I checked
- the page makes **no claim about whether a key already exists**, with the reason stated in the UI copy: Polaris keeps no recoverable plaintext. That is the honest version. Showing "a key exists" from a stored hash is the tempting shortcut and it teaches users the wrong model of what the server holds
- "仅本次显示 / shown once" is stated at the field rather than left implicit

And the guard test bites: seeding `apiKey` state with `'pol_dl_LEAKED'` instead of `null` reds both tests in that file. An assertion that a secret is *absent* is worthless if nothing could have made it present, so I checked that specifically.

PR #547, opened 2026-08-30
feat(admin): manage document processing runtime settings

### LINKED ISSUE #546: feat(admin): manage document processing runtime settings
## Summary

Add persistent administrator controls for document-processing runtime behavior in Polaris.

## Scope

- Persist MinerU enablement, endpoint, timeout, polling interval, retry count, concurrency, and PyMuPDF fallback policy.
- Encrypt and manage a multi-key MinerU pool with masked display, rotation, enable/disable, deletion, and health state persistence.
- Test MinerU connectivity through the same configured runtime settings.
- Resolve database settings over environment defaults while preserving environment fallback until an explicit key pool is configured.
- Apply the runtime policy to the paper-content parsing lifecycle, including MinerU-disabled and fallback-disabled terminal states.

This change is backend-only. It does not include the administrator UI, search behavior, or sensitive configuration values.

## Validation

- 24 document-processing/MinerU/paper-content tests passed.
- 23 related literature, structured-reading, PDF-asset, and evidence tests passed.
- 218 backend tests passed; one unrelated Windows `cp936`/UTF-8 Alembic fixture failure remains in the existing test suite.
- Ruff checks for changed files and `git diff --check` passed.

yyy-OPS:
Closes #546

## Summary

This PR adds persistent administrator controls for Polaris document-processing runtime behavior.

## What changed

- Added database-backed MinerU runtime settings: enablement, endpoint, timeout, polling interval, retries, concurrency, and PyMuPDF fallback policy.
- Added encrypted multi-key MinerU pool management with masked values, rotation, enable/disable, deletion, and persisted health state.
- Added MinerU connectivity testing using the effective runtime configuration.
- Added explicit precedence rules: database configuration overrides environment defaults; environment key fallback remains available until an administrator explicitly establishes a key pool.
- Wired the effective policy into the paper-content parsing lifecycle, including MinerU-disabled and fallback-disabled terminal behavior.

## Scope and compatibility

- Backend only; administrator UI and literature-search behavior are intentionally out of scope.
- Reuses the existing `system_settings` JSON KV storage; no migration is required.
- No secrets or environment values are included.

## Validation

- 24 document-processing/MinerU/paper-content tests passed.
- 23 related literature, structured-reading, PDF-asset, and evidence tests passed.
- 218 backend tests passed; one unrelated existing Windows `cp936`/UTF-8 Alembic fixture failure remains.
- Ruff checks for changed files and `git diff --check` passed.

### DIFF
--- a/src/backend/app/api/admin_settings.py
+++ b/src/backend/app/api/admin_settings.py
@@ -9,6 +9,12 @@ from app.schemas.admin_settings import (
     AffiliationModeRead,
     AffiliationModeUpdate,
     DailyEmbedBackfillResult,
+    DocumentProcessingCredentialCreate,
+    DocumentProcessingCredentialStatus,
+    DocumentProcessingCredentialUpdate,
+    DocumentProcessingProviderTestResult,
+    DocumentProcessingSettings,
+    DocumentProcessingSettingsUpdate,
     EmbeddingSpaceAdoptResult,
     EmbeddingSpaceItem,
     EmbeddingSpaceStatus,
@@ -28,6 +34,7 @@ from app.schemas.admin_settings import (
 from app.schemas.tts import TTSAdminSettings, TTSTestResult, TTSVoicesResult
 from app.services import affiliations as affiliations_service
 from app.services import daily_feed as daily_service
+from app.services import document_processing_settings as document_processing_settings_service
 from app.services import embedding as embedding_service
 from app.services import experiment_settings as experiment_settings_service
 from app.services import lab as lab_service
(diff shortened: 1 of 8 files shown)


yyy-OPS:
Additional local verification on the PR branch: the focused document-processing/MinerU/paper-content suite passes 24 tests in the project virtual environment (72.14s). `git diff --check` and Ruff checks for changed files pass. The system Python environment lacks project dependencies, so the result above was obtained with the repository virtual environment.

tricktreat:
Merged as-is. Based on current main, no migration, clean merge; 34 passed across document-processing settings, MinerU, paper content, literature settings and migrations; ruff clean.

The part I checked hardest is the environment-fallback precedence, because it is the same shape as the credential-pool bug found in #517/#526 — and this gets it right:

```python
if "mineru_credentials" in value:      # keyed on the key being present
    tokens = [... for entry in _credential_pool(value) if entry["enabled"]]
else:
    tokens = [...environment...]
```

Keyed on **presence**, not truthiness. So an administrator who declares a pool and then disables every credential in it gets an empty token tuple rather than a silent slide back onto the environment key — which is precisely the failure where disabling something appears to work and does not.

The test suite pins it, including the name (`test_runtime_keeps_environment_tokens_until_database_pool_is_declared`) and the all-disabled case returning `()`. And it bites: changing the condition to `if _credential_pool(value):` reds `test_mineru_credential_crud_is_masked_encrypted_and_controls_runtime`.

Also good: `_base_url` rejects a URL carrying credentials or a query string, so an endpoint cannot smuggle a key into what looks like configuration.

PR #552, opened 2026-09-02
feat(extension): send library papers and track archive batches

### LINKED ISSUE #545: feat(extension): send library papers and track archive batches
## Problem

The discovery workspace can create one extension batch for selected promoted hits, but papers already present in a library have no equivalent batch action or durable status view. Users also need to see which papers were skipped because a verified PDF already exists and which archive items still require attention.

## Proposed scope

- Add a manager-only extension action to a library paper detail view and the existing multi-select toolbar.
- Create one backend batch per user action, with one independently bound item per `library_id + paper_id` pair.
- Exclude papers with a verified, readable PDF from the browser download payload while preserving each server-side item and binding.
- Dispatch the batch through the shared Polaris extension bridge.
- Show recent batches in a compact dialog with item counts, cached/skipped/uploaded/failed states, and retry guidance.
- Keep public-library visitors read-only.
- Link the empty or disconnected state to `/settings?tab=extension`.

## Dependencies

Implementation should start from the then-current `main` after #533 lands so it reuses the accepted batch types and page bridge. Asset-state controls should reuse #537 after merge. The implementation PR must not duplicate those helpers or stack their branches.

## Non-goals

- publisher-specific automation;
- native fixed-directory access;
- changing PDF sharing or grant rules;
- treating a batch position as a paper identity;
- downloading a PDF that already has a verified readable asset.

## Acceptance criteria

- Selecting multiple eligible papers creates one batch with multiple items.
- Skipped PDFs do not enter the browser download payload and do not shift another paper's archive binding.
- A PDF can be archived independently of the batch order or task index.
- Managers see clear dispatch and archive progress; read-only visitors see no mutation controls.
- Frontend tests, extension bridge tests, the production build, and mobile/desktop visual checks pass.

yyy-OPS:
## Summary

- add library paper actions for sending one paper or a multi-selection to the Polaris extension
- create one download batch per user action while preserving an independent library/paper binding for every item
- skip duplicate downloads for papers that already have a readable PDF while retaining their server-side batch records
- add a polling modal for recent extension batches and enforce library membership on batch creation and archive upload

## Validation

- Backend targeted tests: 12 passed
- Frontend tests: 119 passed
- Frontend production build: passed
- Ruff: passed

Closes #545

### DIFF
--- a/src/backend/app/api/download_client.py
+++ b/src/backend/app/api/download_client.py
@@ -16,7 +16,7 @@ from app.core.config import get_settings
 from app.core.db import get_session
 from app.core.queue import TaskQueue, get_task_queue
 from app.models.download_client import DownloadApiKey, DownloadBatch, DownloadBatchItem
-from app.models.library_direction import DirectionLibrary
+from app.models.library_direction import DirectionLibrary, LibraryPaper
 from app.models.paper import Paper
 from app.models.user import User
 from app.schemas.download_client import (
@@ -40,6 +40,7 @@ router = APIRouter(tags=["download-client"])
 SCOPES = ["download_batch:read", "download_batch:update", "paper_pdf:upload"]
 LEASE_MINUTES = 10
 TERMINAL_ITEM_STATUSES = {"uploaded", "skipped", "failed", "blocked", "cancelled"}
+DOWNLOADABLE_LIBRARY_STATUSES = ("scored", "fetched", "compiled", "included")
 
 
 def _hash_key(value: str) -> str:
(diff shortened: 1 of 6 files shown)


tricktreat:
Landed as #560 (merged) with authorship preserved, plus one fix and two tests.

`_require_library_paper` is called in three places. Removing each in turn:

| call site | suite with it removed |
|---|---|
| create | 1 failed |
| archive | 1 failed |
| **upload** | **6 passed** |

The upload check had no test, and it is the one I would least want unguarded — upload is what the extension reaches with only an API key and a lease, so a paper removed from the library (or reverted to `candidate`) after the batch was created could still have a PDF written against it, silently. Added a test that claims an item, reverts membership, uploads, and expects 404.

Separately, the UI offered the action for papers the backend rejects: `PapersTab` lists candidates under the `visible` filter and the button was gated only on the in-flight flag, so sending a candidate produced a bare `LIBRARY_PAPER_NOT_FOUND`. Added `canSendToExtension` mirroring `DOWNLOADABLE_LIBRARY_STATUSES`, gated the button, and put the reason in the tooltip.

Excluding `candidate` from the downloadable set is the right call and consistent with the unpromoted-hits gate — the gap was only that the UI did not know about it.

PR #553, opened 2026-09-02
feat(reading): resolve AI citations to authorized full-text evidence

### LINKED ISSUE #544: feat(reading): resolve AI citations to authorized full-text evidence
## Problem

Polaris can persist versioned evidence anchors and serve authorized structured content, but generated citations still lack one complete frontend path from a rendered reference to the cited sentence. Opening only the paper or choosing the first matching sentence is too coarse, especially when the same sentence appears in the abstract and full text.

## Proposed scope

- Render evidence-aware citations in paper Wiki content and other text surfaces that have an evidence manifest.
- Carry the library, paper, content version, and anchor identity in the navigation contract without exposing restricted text in the URL.
- Resolve the anchor through the library-authorized evidence endpoint.
- Open the PDF reader first and highlight the returned page rectangles or a quote match in the PDF text layer.
- Use structured content only as a fallback when the PDF text layer cannot resolve the quote.
- Fall back to the paper page when a reparse or permission change makes the fine-grained anchor unavailable.
- Distinguish stale anchors, temporary resolution failures, and authorization failures in the UI.
- Never choose an ambiguous duplicate quote without contextual evidence.

## Dependencies

Implementation should start from the then-current `main` after #537 lands. Workflow surfaces should consume the evidence manifests introduced by #531 after that contract is merged. The implementation PR must not stack either branch or copy their files into a parallel implementation.

## Non-goals

- changing the evidence persistence model;
- making private full text globally visible;
- treating paragraph boundaries as stable across reparses;
- emitting sentence citations when the generation has only metadata or an abstract.

## Acceptance criteria

- A valid citation opens and highlights the intended PDF sentence when coordinates or a confident quote match are available.
- Repeated text is disambiguated with chunk, section, and neighboring-text hints.
- A stale citation falls back to the paper instead of producing a dead link.
- Revoked access cannot resolve or display restricted evidence.
- MinerU and PyMuPDF content paths both pass focused tests.
- Desktop and mobile reader checks show no layout regression.

yyy-OPS:
## Summary

- resolve AI evidence citations through a library-authorized backend endpoint
- prefer PDF text-layer highlighting, with rectangle coordinates as a fallback
- fall back to structured full text when PDF positioning is unavailable
- degrade stale, reparsed, or no-longer-authorized anchors to paper-level navigation instead of dead links
- avoid arbitrary highlighting when repeated text cannot be disambiguated
- wire evidence navigation into the Wiki and paper reading surfaces

## Scope

This PR covers evidence navigation for the Wiki and paper reader surfaces. It does not claim sentence-level provenance across every AI-generated artifact in Polaris.

## Validation

- Backend targeted tests: 31 passed
- Frontend tests: 121 passed
- Frontend production build: passed
- Ruff: passed

Closes #544

### DIFF
--- a/src/backend/app/api/evidence.py
+++ b/src/backend/app/api/evidence.py
@@ -58,4 +58,19 @@ async def resolve_library_evidence(
         session, asset_id=version.asset_id, library_id=library_id
     ) is None:
         raise HTTPException(status.HTTP_404_NOT_FOUND, detail="EVIDENCE_ASSET_NOT_FOUND")
-    return result
+    resolved_chunk = next(
+        (chunk for chunk in chunks if chunk.id == result.chunk_id),
+        None,
+    )
+    anchor_query = f"?library_id={library_id}&evidence={anchor.id}"
+    return result.model_copy(
+        update={
+            "library_id": library_id,
+            "content_version_id": version.id,
+            "parser": version.parser,
+            "section_path": (
+                resolved_chunk.section_path or [] if resolved_chunk is not None else []
+            ),
+            "href": f"/papers/{paper_id}/read{anchor_query}",
+        }
+    )
--- a/src/backend/app/schemas/evidence.py
+++ b/src/backend/app/schemas/evidence.py
@@ -32,7 +32,9 @@ class EvidenceResolution(BaseModel):
     model_config = ConfigDict(frozen=True)
 
     paper_id: uuid.UUID
+    library_id: uuid.UUID | None = None
     anchor_id: uuid.UUID | None
+    content_version_id: uuid.UUID | None = None
     status: EvidenceFallback
     anchor_type: EvidenceAnchorType
     quoted_text: str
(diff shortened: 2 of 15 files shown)


tricktreat:
Merged as-is. Backend 18 passed, frontend `tsc --noEmit` clean and `vitest run` 124 passed, ruff clean, no migration.

Two things I checked rather than took on trust:

**"Dead links" are impossible by construction, not by handling.** The citation href is always `/papers/{paperId}/read`, with `library_id`/`evidence`/`content_version_id` attached only as optional query params when a unique ref and a library are present. So an anchor that later becomes stale, reparsed, or unauthorized costs you the highlight, never the link. That is a better shape than resolving first and rendering a link conditionally — there is no failure path where the user gets a broken target.

**The ambiguity guard exists on both sides and both bite.** Frontend `matchingReference` returns null unless `matches.length === 1`; backend `resolve` requires `len(matches) == 1`. Relaxing either — `matches[0] ?? null` on the frontend, `>= 1` on the backend — reds the corresponding test. Highlighting an arbitrary occurrence of a repeated sentence is the kind of wrong that looks right, so guarding it in both places and testing both is the correct amount of paranoia.

The scope note in the PR body is worth keeping — being explicit that this covers Wiki and reader surfaces and does not claim sentence-level provenance everywhere is more useful than a claim that would quietly overreach.

PR #557, opened 2026-09-02
feat(admin): add document processing settings UI

### LINKED ISSUE #554: feat(admin): add document processing settings UI
### Problem

The document-processing runtime and credential endpoints are available after #547, but administrators cannot manage them from the web application.

### Scope

- Add a Document processing tab to administrator settings.
- Expose MinerU API root, timeout, polling, retry, concurrency, and parser fallback settings.
- Add write-only MinerU credential create, edit, enable, delete, and connection-test actions.
- Keep credential changes immediate and general settings explicit.

### Acceptance criteria

- Secrets are never read back into the browser.
- Invalid parser policies and numeric bounds are rejected before submission.
- The page follows the existing Polaris settings layout.
- Frontend tests and production build pass.

yyy-OPS:
## Summary

- add a Document processing tab to administrator settings
- expose the runtime policy added in #547
- manage the write-only MinerU credential pool through add, edit, enable, delete, and test actions
- validate parser policy and worker limits before saving

The branch is based directly on main at a330d15. It does not depend on the other frontend completion pull requests.

## Tests

- npm test -- --run: 121 passed
- npm run build
- focused document-processing settings tests: 4 passed

Closes #554

### DIFF
--- a/docs/superpowers/specs/2026-09-02-polaris-upstream-frontend-completion-design.md
+++ b/docs/superpowers/specs/2026-09-02-polaris-upstream-frontend-completion-design.md
@@ -1,0 +1,44 @@
+# Polaris upstream frontend completion
+
+## Scope
+
+This change set closes three frontend gaps left after the literature backend was merged:
+
+1. expose document-processing runtime settings and the MinerU credential pool to administrators;
+2. make the existing Unpaywall resolver selectable and testable from literature settings;
+3. keep the literature list/detail workspace usable when the main content area is 1025-1280 pixels wide.
+
+Each item is delivered as an independent pull request based on the same upstream commit. None of the branches depends on another branch.
+
+## Document-processing administration
+
+The administrator page gains a `Document processing` tab. It calls the existing `/api/admin/settings/document-processing` endpoints and does not introduce a second configuration store.
+
+The page contains two bounded sections:
+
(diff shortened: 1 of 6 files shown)


tricktreat:
Landed as #561 (merged) with authorship preserved, plus the credential test strengthened.

The stripping test is real here — the fixture actually carries `mineru_credentials`, so removing the destructure reds it. That is a step up from the equivalent in #535, which asserted against a draft that never had the field and stayed green when the allowlist was replaced by a spread.

Two small strengthenings: the fixture's credential list was `[]`, so I filled it with a full `DocumentProcessingCredentialStatus` carrying a masked preview, and added an assertion that `••••1234` never appears in the serialised draft. Absence-of-key and absence-of-value are different failures.

One shape note I did not change: the stripping is a destructure-exclusion, i.e. a denylist of one. `buildLiteratureSettingsUpdate` on the sibling settings page uses an explicit allowlist, which fails closed when the response grows a field. If a masked health or preview field is ever added to `DocumentProcessingSettings`, this one flows it straight back. Not a defect today, just the difference between the two pages.

Also worth noting positively: the client validation mirrors the backend `_base_url` rule exactly, rejecting a MinerU root carrying credentials, a query or a fragment — so a key cannot be smuggled through the endpoint field, and the user finds out before the round trip.

PR #558, opened 2026-09-02
feat(literature): expose Unpaywall resolver health

### LINKED ISSUE #555: feat(literature): expose Unpaywall OA resolver health
### Problem

Unpaywall is used to resolve OA PDF locations during candidate caching, but administrators cannot see or test that integration.

### Scope

- Show Unpaywall separately from bibliographic search providers.
- Probe the resolver with a DOI through the existing provider health endpoint.
- Persist the resulting health state with other literature provider checks.
- Keep Unpaywall out of candidate search quotas and source snapshots.

### Acceptance criteria

- A resolver test makes a real DOI lookup.
- Resolver failures return structured health feedback.
- Unpaywall never appears as a zero-result bibliographic database.
- Backend and frontend tests pass.

yyy-OPS:
## Summary

- show Unpaywall as an OA resolver instead of a bibliographic search source
- test the configured resolver with a real DOI lookup
- persist resolver health with the existing literature provider health records
- keep resolver calls out of candidate search quotas and source snapshots

The branch is based directly on main at a330d15. It does not depend on the document-processing UI or responsive layout pull requests.

## Tests

- backend administrator settings and discovery runtime: 30 passed
- Ruff checks passed
- npm test -- --run: 118 passed
- npm run build

Closes #555

### DIFF
--- a/src/backend/app/api/admin_settings.py
+++ b/src/backend/app/api/admin_settings.py
@@ -242,6 +242,27 @@ async def test_literature_provider(
     source = payload.source.strip().lower()
     started = time.perf_counter()
     try:
+        if source == "unpaywall":
+            from app.services.literature.multi_source import MultiSourceClient
+
+            resolver = MultiSourceClient()
+            try:
+                matched = await resolver.lookup_unpaywall(payload.query)
+            finally:
+                await resolver.aclose()
+            if matched is None:
+                raise ValueError("resolver returned no record; check the contact email and DOI")
+            result = LiteratureProviderTestResult(
+                source=source,
+                ok=True,
+                latency_ms=round((time.perf_counter() - started) * 1000),
+                fetched_count=1,
+                detail="OA resolver responded",
+            )
+            await literature_settings_service.record_provider_health(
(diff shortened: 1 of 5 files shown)


tricktreat:
Landed as #562 (merged) with authorship preserved, plus one test.

The modelling is right — `resolverOnly` sits alongside the existing `metricOnly` flag instead of introducing a parallel mechanism, and all three derived lists come off the one source table. The backend probe is a genuine resolver probe against a real DOI, and its test name states what it defends.

But the frontend half of that claim had no guard: removing `!source.resolverOnly` from the `SEARCH_SOURCES` filter left all six tests green. Unpaywall drifting back into the search list would get queried as a bibliographic source — consuming candidate quota, entering source snapshots, returning no papers — and from the UI that reads as "this source found nothing" rather than "this source should never have been asked".

Added a test pinning the separation in both directions: resolver-only and metric-only providers stay out of `SEARCH_SOURCES`, `RESOLVER_SOURCES` is exactly `['unpaywall']`, and real search sources stay in — so mislabelling a search source as resolver-only fails too.

PR #559, opened 2026-09-02
fix(frontend): adapt literature workspaces to constrained widths

### LINKED ISSUE #556: fix(frontend): adapt literature workspaces to constrained main widths
### Problem

At 1025 to 1280 pixel viewports, or when a docked panel narrows the main area, literature navigation labels collapse and list/detail panes become too narrow.

### Scope

- Use the existing main-area container width as the responsive signal.
- Stack literature list/detail panes when the available content width is at most 1180 pixels.
- Keep literature tabs on one line with horizontal scrolling.
- Apply the behavior to personal, project, and public literature workspaces.

### Acceptance criteria

- Navigation labels do not wrap character by character.
- Detail titles remain readable.
- The page has no document-level horizontal overflow.
- Existing mobile behavior remains unchanged.

yyy-OPS:
## Summary

- trigger literature workspace reflow from the actual main-area width
- stack list and detail panes below 1180 pixels of available content
- keep navigation labels on one line with horizontal scrolling
- cover personal, project, and public literature libraries

The branch is based directly on main at a330d15 and changes no business logic. It does not depend on the other frontend completion pull requests.

## Tests

- npm test -- --run: 119 passed
- npm run build
- visual checks at 1180px and 1025px main viewports
- verified zero document-level horizontal overflow at 1025px

Closes #556

### DIFF
--- a/src/frontend/src/features/libraries/LibraryBrowse.tsx
+++ b/src/frontend/src/features/libraries/LibraryBrowse.tsx
@@ -925,7 +925,7 @@ export function LibraryBrowse({ libraryId }: { libraryId: string }) {
 
   return (
     <>
-      <div className="row" style={{ marginBottom: 14, justifyContent: 'space-between' }}>
+      <div className="row literature-workspace-tabs" style={{ marginBottom: 14, justifyContent: 'space-between' }}>
         <Segmented<BrowseTab>
           options={[
             { v: 'discover', label: tr('发现文献', 'Discover') },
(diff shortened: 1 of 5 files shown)


tricktreat:
Landed as #563 (merged) with authorship preserved, plus one test.

Keying the reflow to the main area rather than the viewport is the right call, and the CSS comment states the reason better than a PR description could — the Buddy dock narrows the main area without changing the viewport, so viewport rules keep believing there is room. That is a failure mode this repo has hit before.

One guard added: a container query naming a container that does not exist **never applies and never errors** — layout stays wide, console stays clean, tests stay green. The existing test pins the query side (`@container mainarea (max-width: 1180px)`) but nothing on the declaration side, so deleting `container-name: mainarea` was invisible. Now both properties are asserted, plus that every `@container` rule names a container that is actually declared.

The source-text testing style is the right tool here — container queries cannot be exercised in jsdom, so behavioural assertions would be theatre. Pinning the text is honest about what is being checked.

PR #779, opened 2026-09-14
A library can be given a discipline from the UI

### LINKED ISSUE #775: One discipline pack does not show the format generalises
One discipline pack ships today — structural engineering. It proves a pack can be *loaded*; it does not prove the format can describe a second field. With a single example, any assumption baked into the format that happens to suit structural engineering looks like a general rule, and the first person to write a pack for their own field discovers otherwise.

The pack format is the whole plugin story for literature: it is what lets a non-CS researcher replace the built-in `method@1` card — `purpose / mechanism / baseline / dataset / protocol`, a machine-learning shape — without touching the repository. "Works for many disciplines" is currently an untested claim.

## What a second and third pack would actually test

Three fields that stress different parts of the format:

- **Wet-lab biology** — the method is *system → perturbation → readout → controls*. Nothing in common with structural engineering except the two cross-disciplinary retrieval axes. Reproducibility in life sciences rests almost entirely on the control setup, so it needs to be its own field rather than prose buried in a method description.
- **Synthetic chemistry / materials** — conclusions are nearly always *"metric X reached value Y under condition Z"*. That needs the `entries` field kind, which no shipped pack currently exercises. Flattened into prose, a question like *"which of these achieved over 80% yield"* degrades to full-text keyword luck.
- **Clinical / epidemiology** — the field already has a controlled vocabulary (PICO). This tests whether the format can carry a convention that *already exists* rather than forcing it into a different shape.

## Checks that should apply to every pack, not one

The current tests assert things about the structural pack by name. A new pack that parses gets no further scrutiny — including the rule that a `method` card must keep `purpose` and `mechanism`, whose absence has no error, just a discipline whose cards are extracted, stored, and never retrievable.

The suite should walk the pack directory instead, so a pack added later is checked automatically: names unique, retrieval axes present, no pack shipping the ML axes it exists to replace, `entries` fields declaring their keys, packs staying scoped so a library only pays for its own.

Worth asserting explicitly that the packs' domain fields are **disjoint** from each other. If two packs share most of their fields, the format has only ever described one discipline in several costumes.

## Known format gap

Field kinds are `text | list | entries` — there is no numeric kind. A yield, a sample size, a temperature all have to be stored as text, which works for display but cannot be sorted or filtered numerically. Not blocking, and not addressed here.

tricktreat:
Closes #775. Follows #776, which shipped the packs themselves.

Four discipline packs now ship, and until this PR **none of them could be turned on**. `PATCH /libraries/{id}` accepted a `discipline` and the extraction path honoured it, but nothing in the product ever sent one. A researcher in structural engineering, synthesis or clinical medicine still got method cards shaped for machine learning, with no indication that a better fit existed or how to ask for it.

## Listing has to come from disk

`GET /disciplines` scans the pack directories on every call rather than serving a cached list, and the frontend asks it rather than shipping its own copy of the names. Packs are data: a user drops a YAML into `<data_dir>/disciplines/` and it should appear in the picker. A hardcoded list means someone follows the documentation, writes their own pack, and then cannot find it in the UI — which is precisely the failure the file-over-app arrangement exists to avoid.

Read-only and plain authentication, not owner: knowing which disciplines are installed carries no privilege, and the person choosing one is an ordinary user.

A broken pack is skipped rather than raising, so one bad indentation in a hand-written file cannot take the listing down for everyone — a test drives exactly that.

## The picker

A card on the library governance page, next to inclusion settings. It states plainly what changes — the default card is shaped for machine learning (purpose, mechanism, baseline, dataset); choosing a discipline swaps in that field's own — and that it applies to papers extracted from then on, not retroactively.

Two states are called out rather than left to be inferred:

- **No packs installed** — says where to put one instead of showing an empty dropdown.
- **A pack with zero schemas** — says that choosing it changes nothing. Silently accepting the selection would let someone believe their discipline was applied.

`disciplinePatchValue` is extracted and unit-tested because the empty case is a real trap: the "general" option must send `null`, not `""`. An empty string is compared against installed pack names, matches nothing, and is rejected as an unknown discipline — so clearing the discipline would fail to save, citing a value the user never typed.

## Verification

**35 passed** on the backend across the new endpoint, pack and library-discipline suites; `ruff` clean. Frontend: `tsc` clean, **186 passed**, production build succeeds.


tricktreat:
Self-review addition in `31254290`: a library can hold a discipline whose pack has since been removed from the data directory.

The select had no matching option for that value, so it rendered blank — indistinguishable from *General* — while the library still carried the name and extraction had quietly fallen back to the builtin fields. The picker now keeps the stored value as an explicit option marked as no longer installed, and says that extraction has fallen back, with the two ways out.

Also filled in `discipline` in `update_library`'s docstring, which still listed only the fields that predated it.

PR #780, opened 2026-09-14
PubMed can fill the daily pool, and you can subscribe to it

### LINKED ISSUE #778: arXiv is still the only source that can fill the daily pool
The daily pool is now source-agnostic in its plumbing — subscriptions carry a source, the fetch loop asks the registry which sources can supply daily increments, and #773 makes several sources per term merge rather than overwrite. None of that helps yet, because **`arxiv` is the only adapter that implements `fetch_new`**. A biologist, a chemist or a clinician subscribing to their own field still gets an empty pool, and the daily feed remains arXiv in practice however source-neutral the code reads.

Eleven sources are registered (openalex, semantic, arxiv, pubmed, crossref, europepmc, hal, core, base, sciverse, unpaywall). The ones that matter for non-CS fields — PubMed for life sciences, Europe PMC for biomedicine, Crossref for chemistry and engineering journals — all go through `MultiSourceAdapter`, which implements `search` and nothing else.

## Three concrete obstacles

**1. The search contract has no day granularity.** `SourceSearchRequest` carries `start_year` / `end_year` only. "Papers that appeared today" cannot be expressed through the path every multi-source provider already uses, so `fetch_new` cannot simply be a thin wrapper over `search`.

**2. Record shape.** `fetch_new` returns **source-native** records by design — `upsert_entries` reads `title`, `arxiv_id`, `doi`, `year`, `authors`, `announce_type` off them — whereas `MultiSourceAdapter.search` returns normalized `LiteratureCandidate` objects. A mapping belongs somewhere, and putting it in the wrong place would push normalization into the daily path, which the source protocol deliberately keeps out.

**3. Indexing lag is the normal case.** arXiv announces a batch and declares its date; PubMed and Crossref index continuously and with delay, so a strict "published today" query usually returns nothing at all. A window of two or three days plus pool deduplication is the workable shape — re-seeing a paper is already harmless, since `find_pool_paper` dedupes on identity and `upsert_entries` merges into the existing entry without touching `feed_date`.

The batch-date machinery already anticipates this: `fetch_new` may return `batch_at=None`, which skips the stale check and falls back to the run date. A source that genuinely has no announcement batch should return `None` rather than inventing one.

## Why this is the blocking piece

`#775` ships packs for biology, chemistry and clinical research, and a library can now declare one. But a discipline pack only changes how papers are *read*. Where the papers *come from* is still one preprint server for one group of fields. A researcher outside CS can now get their own method-card fields applied to a pool that never fills.

## Related wording

`daily.fetch` still calls subscriptions "arXiv 分类" in its user-facing notes and hardcodes arXiv's ~04:00 UTC announcement schedule in its advice. Accurate today, wrong the moment a second source lands — worth fixing in whatever change adds one.

tricktreat:
Closes #778.

The daily pool's plumbing has been source-agnostic for a while, but `arxiv` was the only adapter implementing `fetch_new`. For anyone outside CS the pool was simply empty, and no amount of source-neutral code changed that. This adds the second source and the entrance to subscribe to it — one without the other would be another capability nobody can reach.

## PubMed, by index date rather than publication date

`fetch_new_pubmed` queries `[edat]` (the date a record entered PubMed), not `[pdat]` (the date printed on the paper). A paper published in December can enter PubMed in March; asking for "today" by publication date would miss it permanently. Index date is the real analogue of arXiv's *"what was announced today"*.

The window looks back two days rather than one. PubMed indexes continuously, so a run just after midnight would see almost nothing, and the daily pool is the only supply every library has — a missed day cannot be recovered. Re-seeing yesterday's records is free: `find_pool_paper` deduplicates on identity and `upsert_entries` merges into the existing entry without touching `feed_date`.

Results are sorted by date, not relevance: within a fresh window, relevance ordering would let older records crowd out the new ones the pool is asking for.

`fetch_new` returns `batch_at=None`. PubMed has no announced batch, so declaring a date would be invention — and the caller uses that to skip the "fetched too early" check, which does not apply to a rolling index. The daily path already anticipated sources like this.

## The capability lives on PubMed, not on the shared adapter

`PubMedAdapter` subclasses `MultiSourceAdapter` rather than `fetch_new` being added to the base. Capability detection is `hasattr`, so putting it on the base would have crossref, hal, core and the rest all claiming they can supply daily papers — then failing once per subscribed term, every day.

The three-step esearch → esummary → efetch assembly is factored into `_pubmed_rows`, shared by search and daily. Duplicating it would mean a change to PubMed's fields or key parameters silently applying to only one of the two.

## Subscribing to it

`daily.subscriptions` could previously only be written from the service layer — so even a source that *could* supply daily papers had no way to be subscribed. `GET`/`PUT /daily/subscriptions` exposes the full multi-source view, and a settings section lets non-arXiv sources be subscribed by search term.

The available sources come from the capability probe rather than a hardcoded list, in the API and in the UI both: installing a source that can do daily should make it subscribable immediately, and removing one should show up immediately too.

Subscribing to a source that *cannot* currently supply daily papers is **flagged, not refused** — a source may simply be missing an API key or be temporarily unavailable, and discarding the subscription would make someone retype it. `supports_daily: false` comes back on read and the UI says so in place, which is also the only way to explain an empty pool.

`dailySubscriptionPayload` is extracted and tested because `PUT` is a whole-list replacement: sending only the section being edited would wipe the arXiv categories managed by the section above it. That failure surfaces the same day as an empty pool, to someone who only added a search term.

## Wording that became wrong

`daily.fetch` told users with no subscriptions to add "arXiv 分类". With a second source that hides an entrance that now exists, so it names both.

## Verification

**90 passed** across the PubMed, subscription-API, source-registry, daily-feed and golden suites; `ruff` clean; both golden chains byte-identical. Frontend: `tsc` clean, **187 passed**, production build succeeds.

No network in the tests — `_pubmed_rows` is stubbed, and the assertions are on the query that would be sent: that it uses `[edat]` and not `[pdat]`, that the window spans more than a day, and that the sort is by date.


tricktreat:
Self-review caught a stale-cache bug and it is fixed in `5750a5ed`.

The arXiv categories section invalidated only `['daily-categories']` on save. This section holds the arXiv subscription in its own copy of `['daily-subscriptions']` and sends it back on save, because `PUT` is a whole-list replacement — so editing arXiv categories and then saving a search term here would have written the pre-edit arXiv list back, silently reverting the first change.

The arXiv section now invalidates both keys.

Issue #716, opened 2026-09-07
Audit fix 1: wire Runner v2 dispatch, credentials, and leases into the experiment path


tricktreat:
Part of #715 (item 1). The merged v2 backends are unreachable in production:

- the experiment action layer switches from the v1 `open_runner` container-presence dispatch to `resolve_backend(plan)` — `python-ml` keeps byte-identical behavior (its adapter delegates to the same primitives), non-ML backends become reachable
- experiment creation stops hard-requiring `kind=="ssh"`: backends whose manifest declares no ssh credential need (ngspice/openfoam/fmu run on local docker / in-process) accept `credential_id=None`; the ssh requirement stays for python-ml
- `resource_leases.acquire(wait=True)` finally gets its production caller: prepare acquires against the backend's declared resource needs when a `resource_id` is present (recorded since #677), release already hooks the four terminal paths
- existing experiment tests must pass unchanged (python-ml path byte-compatible); new tests: dispatch to a fake v2 backend, credential-optional creation, lease acquired/queued/released across a run
- goldens byte-identical; zero new failures vs #581

Issue #674, opened 2026-09-06
P3: autonomous experiments (Runner v2, process packs, first non-CS backends)


tricktreat:
Successor to #660 (P2.5 complete). Plan: `docs/rfcs/2026-09-07-p3-autonomous-experiments-plan.md` (PR #673).

- [ ] R1 Runner v2 contract + registry + python-ml adapter
- [ ] R2 resources, leases, polymorphic connection credentials
- [ ] R3 process packs phase 1 (format, interpreter, built-in base/experiment pack)
- [ ] R4 zero-license container backends (openfoam, ngspice)
- [ ] R5 FMU backend (FMPy)
- [ ] R6 BYO runner tier 1 (SSH agent formalization)

Waves: R1 → (R2·R3) → (R4·R5) → R6. Engineering exit: an ngspice/openfoam container run and a fixture-FMU run work end to end behind the v2 contract; one built-in pack drives Navigator planning; exclusive resources queue instead of failing. Deferred: session backends with license queueing, instruments, commercial EDA, tier-2 BYO transport.

tricktreat:
## P3 complete (engineering scope)

All six items merged:

- R1 Runner v2 contract, registry, python-ml adapter (#676)
- R2 resources, leases, polymorphic connection credentials (#680)
- R3 process packs phase 1 (#679)
- R4 ngspice + openfoam container backends (#683)
- R5 FMU backend via FMPy (#684)
- R6 BYO runner tier 1 with agent version pinning (#686)

**Engineering exit criteria verified**
- Runner v2 is in place with the existing ML runner unchanged behind it as `python-ml`; dispatch is explicit with no silent fallback.
- A real ngspice container run went through the full v2 lifecycle on this machine: RC-circuit transient, `vmax = 4.908 V` against the theoretical 4.908 V, 529-point metric series, result files registered, exit 0. The openfoam smoke ships double-gated (image is GB-scale); the FMU path runs validate/collect/worker for real with the e2e case honestly skipped on arm64 (x86-only reference binaries).
- `base/experiment` reproduces the hard-coded plan function byte-for-byte across 5 param sets; pack-less runs take the original code path.
- Exclusive resources cannot be double-leased; a second acquirer queues via `wait_and_acquire`; leases release on all four terminal paths.

Also banked along the way: a real key-material exposure channel (connection-exception text relayed into API details) found by the R6 audit and fixed with `scrub_secrets` plus a pinning test.

[...]

PR #783, opened 2026-09-14
Pick the execution backend and process pack when creating an experiment

tricktreat:
Part of #674 — the entrance only, not the rest of P3.

Four runner backends are registered (`python-ml`, `ngspice`, `openfoam`, `fmu`) and process packs can be authored into `<data_dir>/packs/`. **None of them could be selected from the product.** Both dispatch keys carried a comment saying the UI was deferred until non-Python backends landed; those landed, and nothing followed. Every experiment ran `python-ml` on the default plan path, and the other three were reachable only by editing the repository.

## The frontend was the last thing blocking container backends

`#716` already relaxed credentials per manifest: a backend whose `credential_kinds` is empty takes no connection credential, and `ngspice`, `openfoam` and `fmu` are all in that group. But the creation form hard-required one:

```ts
const canSubmit = !!ideaId && !!credentialId && !mutation.isPending;
```

So even with the API willing, a container backend could not be created. That requirement now follows the chosen backend's manifest, and the credential field says *why* it is greyed out rather than simply going quiet.

`backendNeedsSsh` is extracted and tested because its default matters: while the backend list has not loaded, it must answer **true**. Defaulting the other way means a slow list lets someone submit a `python-ml` experiment with no credential and discover it at run time — after the idea and the intake answers are already filled in.

## The list comes from the registries

`GET /experiment-backends` and `GET /experiment-backends/process-packs` read the registries on each call. A hardcoded frontend list would mean a user authoring their own process pack into the data directory never sees it — the same failure the file-over-app arrangement exists to avoid, and the same one #775 and #778 hit.

Backends report more than an id, because a few facts have to be known *before* choosing: whether an SSH credential is needed, whether a licence seat is required, and whether the backend drives real hardware. Finding that out after submitting is expensive when a run takes tens of minutes. Manifests come from the registry's no-argument probe instances, so listing has no side effects — a test asserts two calls agree, since a listing that connected over SSH or started a container would hang the creation form.

A malformed user-authored pack is skipped rather than raising: one bad indentation should not stop everyone from picking a process.

## Existing behaviour is byte-identical

Neither key is sent when nothing is chosen, so an unmodified form produces exactly today's request: default backend, regular planning path.

## Verification

**57 passed** across the new endpoint, runner v2, process-pack and golden suites; `ruff` clean; both golden chains byte-identical. Frontend: `tsc` clean, **188 passed**, production build succeeds.

## Still open in #674

This makes the backends reachable. It does not make the CAD→CAE pipeline work — that needs per-phase backend selection, and the adaptive loop and resource pre-checks this issue otherwise tracks. Sequencing those is a decision, not something to settle by whoever picks it up first.


tricktreat:
Self-review caught a defect this PR introduced, fixed in `7661e8e2`.

The form always sent `credential_id: credentialId`. That was safe while a credential was mandatory — it could never be empty at submit time. Relaxing the requirement for container backends made the empty string reachable, and `ExperimentCreate.credential_id` is `uuid.UUID | None`, so the first `ngspice` experiment anyone created would have been rejected with a 422 naming a field they had deliberately left blank.

The key is now omitted entirely when nothing is selected, which is what the backend's relaxed validator expects.

Also gave the process-pack list an error state instead of hiding the field when the request fails — silently disappearing is the same shape of problem this PR exists to fix.

PR #791, opened 2026-09-14
Comparison and gap ledger follow the library's discipline

### LINKED ISSUE #790: The comparison table compares papers on axes a discipline library does not use
The paper comparison table is built from a module-level constant derived from the two built-in schemas:

```python
COMPARISON_FIELDS = tuple(
    (field.name, field.kind, schema.id)
    for schema in (SKELETON_SCHEMA, METHOD_SCHEMA)
    for field in schema.fields
)
...
schema_ids = {SKELETON_SCHEMA.id, METHOD_SCHEMA.id}
```

So in a library that declares a discipline, comparing papers side by side shows the machine-learning rows — purpose, mechanism, baseline, dataset, protocol — and **none of the discipline's own**. Structural engineering's `structure`, `actions`, `analysis`, `validation` are not queried and have no rows to appear in. The `baseline` and `dataset` rows read *"未抽取"* for every column, because in that library nothing extracts them.

The table therefore compares papers on axes the field does not use, while the axes it does use are absent. Comparison is exactly the feature where domain fields matter most: *"these three papers all analyse blast loading — what differs is the element type and what each validated against"* is the question, and it is the one the table cannot express.

## It already has what it needs

`build_comparison_table` receives `library_id` and already filters membership on it. The discipline is one lookup away, and `schemas_for(discipline)` returns exactly the schemas that library extracts. The field list can be derived per call instead of frozen at import.

The constant being module-level is what froze it: discipline packs are registered lazily at first use, so anything computed at import time is guaranteed to predate them. Worth checking for the same shape elsewhere — a registry read at import is a registry read before the plugins load.

## Row labels

Rows are keyed by field name, so a discipline field would render as `structure`. #788 adds an optional `label` to `SchemaField` and `PackField` for the same reason on method cards; comparison rows should carry it too.

Third instance of one pattern, after #786 and #788: a consumer that hardcodes the built-in schema's shape, so the pack's own fields never reach the surface the user looks at.

tricktreat:
Closes #790. Stacked on #789 — it adds the `label` this uses.

Two more consumers that hardcoded the built-in schema's shape, so a discipline library never saw its own fields.

## The comparison table compared papers on axes the library does not use

`COMPARISON_FIELDS` was a module-level constant built from `(SKELETON_SCHEMA, METHOD_SCHEMA)`. In a structural-engineering library that meant rows for `baseline` and `dataset` reading *"未抽取"* in every column — nothing in that library extracts them — while `structure`, `actions`, `analysis` and `validation` had no rows at all.

Comparison is where domain fields matter most. *"These three all analyse blast loading — what differs is the element type and what each validated against"* is the question, and the table could not express it.

The constant being module-level is what froze it: discipline packs register lazily at first use, so anything computed at import time is guaranteed to predate them. `comparison_fields(discipline)` is now computed per call, and a test registers a pack *after* first calling it to pin that.

Scope is unchanged — skeleton plus one method card. The gaps schema stays out: it is a list of individually sourced entries, which does not read side by side. And only *one* method card is taken, for the reason #787 established: both cards present means comparing the same thing twice, one of them in the vocabulary the pack exists to replace.

## Picking "more specific" cannot be done by sorting

My first attempt took the alphabetically last method id, assuming a namespaced id sorts after the bare one. It does not — `latecomer.method` < `method`, so "more specific" selected the *built-in* card. The test caught it. Specificity is now computed from the discipline name directly rather than inferred from ordering.

## The gap ledger ignored a discipline's own gap card

`library_gaps` filtered `schema_id == GAPS_SCHEMA.id` exactly. The pack format permits `<pack>.gaps` — one of the pack tests uses `gaps` precisely because non-method schemas are unconstrained — so an author can write a discipline gap card today, have it extracted and stored, and see the ledger ignore it.

It now takes the same union `method_schema_ids` applies, for the same reason: a library holds both papers read through its discipline and cross-disciplinary ones carrying only the built-in card.

Latent rather than live — no shipped pack has a gaps schema — but this is the method-card bug from before packs shipped a method card.

## Labels

`ComparisonRow` carries the pack's `label`. The frontend keeps its own zh/en strings for built-in fields and falls back to the label for anything else, so built-in rows render identically and discipline rows read 结构对象 rather than `structure`. CSV export takes the same path — previously it would have exported machine names.

## Verification

**75 passed** across the comparison, gap-ledger, discipline and pack suites plus both golden chains; `ruff` clean. Frontend: `tsc` clean, **193 passed**, production build succeeds.


tricktreat:
Self-review caught a defect this PR introduced, fixed in `82e84631`.

Making the comparison table discipline-aware means an `entries` field can now reach it — the synthesis pack's `metric / value / condition` and the clinical pack's `measure / effect / interval`. `_render_value` handled `list` and let everything else fall through to `str(raw)`, so those cells would have shown a Python literal — `[{'metric': '产率', 'value': '82%', ...}]` — in the table and in the CSV export.

That kind could not appear before this change: neither built-in schema declares one. Adding the discipline's method card is what made it reachable.

Entries now render as `产率 · 82% · 1 M KOH；过电位 · 270 mV · 10 mA/cm²`, matching what #789 does for the same fields on method cards, with empty keys skipped — the clinical pack deliberately leaves an interval blank when the paper reports none, and that should not leave a stray separator.

